PR #6, opened 2025-01-28
WIP validate allowlisted program instructions from config

gitteri:


kora.toml
```diff
@@ -7,12 +7,40 @@ max_signatures = 10
...
\ No newline at end of file
+]
+disallowed_accounts = []
+# if there is an "allowed_programs" listed above and not in the allowed_instructions, then we will allow all instructions for that program
+allowed_instructions = [
+    { program = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA", instructions = ["[3,...]"] }, # when there's no IDL available
+    { program = "11111111111111111111111111111111", instructions = ["transfer"] }, # when there's an IDL available
+    { program = "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL", instructions = ["*"] }, # Ata Program
+]
+# 1. how do I convert from string to u8 discriminators?
```
gitteri, line 25: I'll clean all this up in a future PR as well as add sane example allowed instructions.

Side comment:
I'm considering enabling a couple smart defaults like "transfers" or "defi" that would automatically build an allowlist for good defaults (USDC/USDT etc transfers, well known defi protocols etc) What do you think? worth it? (can be a follow up PR)

src/common/config.rs
```diff
@@ -3,20 +3,36 @@ use std::{fs, path::Path};
 use toml;
 
 use crate::common::{token::check_valid_tokens, KoraError};
+use crate::common::instructions::{get_program_instruction_configs_with_discriminators, ProgramInstructionConfig, ProgramInstructionConfigWithDiscriminators};
 use solana_client::nonblocking::rpc_client::RpcClient;
 
 #[derive(Debug, Deserialize)]
 pub struct Config {
-    pub validation: ValidationConfig,
+    pub validation: RawValidationConfig,
```
gitteri, line 11: if there's a better rust way to do this lmk. Basically want one version of this struct before validation and another validation as we add additional metadata based on the IDL parsed in the validation step

src/common/instructions.rs
```diff
@@ -0,0 +1,187 @@
...
+    let idl: serde_json::Value = serde_json::from_slice(&s[..])?;
+
+    let mut discriminators = HashMap::new();
+
+    // Get instructions from IDL
+    if let Some(instructions) = idl["instructions"].as_array() {
+        for instruction in instructions {
+            if let Some(name) = instruction["name"].as_str() {
+                // Convert instruction name to snake_case for matching
+                let snake_case = to_snake_case(name);
```
gitteri, line 139: Thinking of recommending all provided instructions in the config to be snake case (or I can just automatically convert) so we fuzzy match instead of requiring exact matches

src/common/instructions.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Parse IDL data
+    // TODO: v0.31 and above versions of anchor do not require the discriminator to be 8 bytes...
+    let mut d: &[u8] = &idl_account.data[8..]; // Skip 8-byte discriminator
+    
+    #[derive(BorshDeserialize)]
+    struct IdlAccountData {
+        authority: Pubkey,
+        data_len: u32,
+        data: Vec<u8>,
+    }
```
ngundotra, line 121: I would recommend to pull this outside the method for code styl

src/common/validation.rs
```diff
@@ -123,6 +126,40 @@ impl TransactionValidator {
...
+                continue;
+            }
+
+            // Check if instruction data matches any allowed discriminator
+            let instruction_data = &instruction.data[..8];  // First 8 bytes are the discriminator
+            let instruction_data_vec = instruction_data.to_vec();
+
+            let is_allowed = program_config.discriminators.iter().any(|(disc, _)| {
+                disc == &instruction_data_vec
+            });
```
ngundotra, line 152: the exact size of the discriminator is only 8 bytes for anchor programs...

Better to check `disc == &instruction.data[..disc.len()]` inside the iterator

src/common/validation.rs
```diff
@@ -231,6 +268,7 @@ mod tests {
             max_allowed_lamports: 1_000_000,
             max_signatures: 10,
             allowed_programs: vec!["11111111111111111111111111111111".to_string()],
+            allowed_program_instructions: vec![],
```
ngundotra, line 271: worth adding example test with a program instruction?

src/common/validation.rs
```diff
@@ -123,6 +126,40 @@ impl TransactionValidator {
...
+                .iter()
+                .find(|config| Pubkey::from_str(&config.program_id).unwrap() == program_id)
+                .ok_or_else(|| KoraError::InvalidTransaction(
+                    format!("Program {} not found in allowed instructions", program_id)
+                ))?;
+
+            // If wildcard "*" is present, allow all instructions for this program
+            if program_config.instructions.contains(&"*".to_string()) {
+                continue;
+            }
```
ngundotra, line 144: I would make "*" a constant

src/common/instructions.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        // Process each instruction - could be raw discriminator or name
+        for instruction in &config.instructions {
+            if instruction == "*" {
+                // Wildcard case - allow all instructions
+                instruction_names.push(instruction.clone());
+                continue;
+            }
+
+            if instruction.starts_with('[') && instruction.ends_with(']') {
```
ngundotra, line 48: Can you add an example or test for these cases?

src/common/instructions.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    for config in program_configs {
+        let mut discriminator_map = HashMap::new();
+        let mut instruction_names = Vec::new();
+
+        // Process each instruction - could be raw discriminator or name
+        for instruction in &config.instructions {
+            if instruction == "*" {
+                // Wildcard case - allow all instructions
+                instruction_names.push(instruction.clone());
```
ngundotra, line 44: For wildcard processing, we actually don't need to store any instructions technically, right?

Because later on, we allow any instruction to go through w/o checking discriminator, right?

ngundotra: Nothing actionable here, just wanted to flag

kora.toml
ngundotra: Yea this would be a huge value-add.

Grab some Raydium, Jupiter, and Orca discriminators

gitteri:
Closing in favor of `https://github.com/pkxro/kora/pull/10` to reduce conflicts.

PR #14, opened 2025-02-12
feat: removed `token_price_info` and get token price from an oracle

Gmin2:


Gmin2:
was not compatible with gibwork branch name , so closing it

Issue #12, opened 2025-02-12
[BOUNTY] Use oracle for token price


gitteri:
### General Relayer Information

This bounty is to enhance the currently in development relayer that mostly adheres to the [Relayer spec](https://docs.google.com/document/d/1lweO5WH12QJaSAu5RG_wUistyk_nFeT6gy1CdvyCEHg/edit?tab=t.0). It is an open source permissionless relayer that anyone can fork and run to support their particular use cases. It is written in Rust and has three crates:

1. Lib: This is where the core logic of the application exists. Most updates should happen here.
2. RPC: Thin RPC layer that specifies the endpoints for the Relayer.
3. CLI: Thin CLI layer to allow for direction interaction with the underlying Relayer functionality through a CLI.

The configuration, kora.toml, lives at the top layer and is processed by lib/config.rs. Specific configuration structs exist for different purposes. For example, “kora” specifies the configuration for the RPC layer including things like a rate limit.

More details about the repo can be found in the readme.md

_All features should be well tested with unit tests and integration tests if appropriate._

### Bounty Feature Request
**Use oracle for token price**

Description:
Currently SignTransactionIfPaid endpoint uses the provided token price to determine the amount of tokens that should be paid to the relayer. Instead the relayer should use an oracle / on-chain price to determine this price

Acceptance Criteria:
- Remove the token_price_info from the SignTransactionIfPaid endpoint
- Get the token price from an oracle instead
- Robust retry for getting the token price
- Hook this price into the existing flow to ensure the relayer is paid enough of the provided token to cover fees
- _Bonus_: use multiple oracles and create an average price while removing any extreme values

pkxro:
Done with #29

PR #19, opened 2025-02-13
Feat: Use oracle for token price - added support for Pyth Orace

0xCipherCoder:
This PR adds Pyth Oracle to get a real-time price from the Pyth network. Added test case and retry mechanism for the changes made.
. 
 This pull request was created for https://app.gib.work/bounties/9429edb4-175f-40c5-bc29-eb57e79b3a71 in an attempt to solve a bounty #12 . Payment for the bounty is immediately sent to the contributor after merge.

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -1,5 +1,5 @@
 use serde::{Deserialize, Serialize};
-use std::{fs, path::Path};
+use std::{collections::HashMap, fs, path::Path};
 use toml;
 
 use solana_client::nonblocking::rpc_client::RpcClient;
@@ -20,6 +20,7 @@ pub struct ValidationConfig {
     pub allowed_tokens: Vec<String>,
     pub allowed_spl_paid_tokens: Vec<String>,
     pub disallowed_accounts: Vec<String>,
+    pub token_symbols: HashMap<String, String>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -57,16 +58,17 @@ mod tests {
     #[test]
     fn test_load_valid_config() {
         let config_content = r#"
-            [validation]
-            max_allowed_lamports = 1000000000
-            max_signatures = 10
-            allowed_programs = ["program1", "program2"]
-            allowed_tokens = ["token1", "token2"]
-            allowed_spl_paid_tokens = ["token3"]
-            disallowed_accounts = ["account1"]
+        [validation]
+        max_allowed_lamports = 1000000000
+        max_signatures = 10
+        allowed_programs = ["program1", "program2"]
+        allowed_tokens = ["token1", "token2"]
+        allowed_spl_paid_tokens = ["token3"]
+        disallowed_accounts = ["account1"]
+        token_symbols = { "So11111111111111111111111111111111111111112" = "SOL", "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v" = "USDC", "Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB" = "USDT" }
 
-            [kora]
-            rate_limit = 100
+        [kora]
+        rate_limit = 100
         "#;
 
         let temp_file = NamedTempFile::new().unwrap();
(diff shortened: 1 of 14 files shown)


0xCipherCoder:
I tried using Switchboard as well to have combined price but there are multiple dependency issues with their Solana SDK.

crates/lib/src/oracle/pyth.rs
```diff
@@ -0,0 +1,173 @@
...
+            "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
+            "Gnt27xtC473ZT2Mw5u8wZ68Z3gULkSTb5DuxJy7eJotD"
+        );
+        // USDT/USD
+        m.insert(
+            "Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB",
+            "3vxLXJqLqF3JG5TCbYycbKWRBbCJQLxQmBGCkyqEEefL"
+        );
+        m
+    };
```
gitteri: Is there a way to get the correct price accounts dynamically instead of hard coding these?

crates/lib/src/transaction/paid_transaction.rs
```diff
@@ -26,14 +27,24 @@ pub async fn sign_transaction_if_paid(
...
 
+    let oracle_client = OracleClient::new(rpc_client);
+
+    let sol_mint = Pubkey::from_str_const(SOL_MINT);
+
+    let token_price_info = oracle_client
+        .get_token_price(&sol_mint)
+        .await
+        .map(|price| TokenPriceInfo { price })
+        .unwrap_or(TokenPriceInfo { price: 0.0 });
```
gitteri: Instead of a 0 price we should return an error

crates/lib/src/oracle/pyth.rs
0xCipherCoder: I am checking this

crates/lib/src/transaction/paid_transaction.rs
0xCipherCoder: Sure will do.

crates/lib/src/oracle/pyth.rs
0xCipherCoder: @gitteri I have added functionality for getting all USDC pairs supported by Pyth and checking for whitelisted token symbols with mint stored in the config. This will help support any token that is whitelisted in the future. Let me know if the approach is fine.

crates/lib/src/transaction/paid_transaction.rs
0xCipherCoder: I am throwing an error now in case of price is not available from Oracle.

0xCipherCoder:
@gitteri I added and updated test cases wherever changes have been made. Please find the screenshots for your reference.
Screenshot 2025-02-14 103359
Screenshot 2025-02-14 103315

crates/lib/src/transaction/paid_transaction.rs
```diff
@@ -27,15 +27,17 @@ pub async fn sign_transaction_if_paid(
     let margin = margin.unwrap_or(0.0);
     let required_lamports = (min_transaction_fee as f64 * (1.0 + margin)) as u64;
 
-    let oracle_client = OracleClient::new(rpc_client);
+    let oracle_client = OracleClient::new(rpc_client, validation).await?;
```
gitteri, line 30: It's much better to create the mapping on server start up rather than calling pyth on every single transaction request. You can update the mapping every few hours to keep things in sync.

crates/lib/src/config.rs
```diff
@@ -57,16 +58,17 @@ mod tests {
...
-            [kora]
-            rate_limit = 100
+        [validation]
+        max_allowed_lamports = 1000000000
+        max_signatures = 10
+        allowed_programs = ["program1", "program2"]
+        allowed_tokens = ["token1", "token2"]
+        allowed_spl_paid_tokens = ["token3"]
+        disallowed_accounts = ["account1"]
+        token_symbols = { "So11111111111111111111111111111111111111112" = "SOL", "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v" = "USDC", "Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB" = "USDT" }
```
gitteri, line 69: You should be able to get the symbol from the token itself instead of relying on manual mapping

Issue #17, opened 2025-02-13
[BOUNTY] Abstract token interactions behind trait


gitteri:
### General Relayer Information
This bounty is to enhance the currently in development relayer that mostly adheres to the [Relayer spec](https://docs.google.com/document/d/1lweO5WH12QJaSAu5RG_wUistyk_nFeT6gy1CdvyCEHg/edit?tab=t.0). It is developed to be an open source permissionless relayer that anyone can fork and run to support their particular use cases. It is written in Rust and has three crates:

Lib: This is where the core logic of the application exists. Most updates should happen here.
RPC: Thin RPC layer that specifies the endpoints for the Relayer.
CLI: Thin CLI layer to allow for direction interaction with the underlying Relayer functionality through a CLI.
The configuration, kora.toml, lives at the top layer and is processed by lib/config.rs. Specific configuration structs exist for different purposes. For example, “kora” specifies the configuration for the RPC layer including things like a rate limit.

_All features should be well tested with unit tests and integration tests if appropriate._

### Bounty Feature Request
Abstract token interactions behind trait

Description:
Note: this is a prerequisite for issue #18. 
Currently the tokenkeg program is used directly throughout the application making it more difficult to add token22 support. To make supporting token22 easier, this bounty is to create a trait that can be used instead. 

Acceptance Criteria:
* Create a token trait that can be used as an interface for token related actions
* Create a tokenkeg implementation of the trait to be used
* Use this new trait for all interactions with tokenkeg instead of directly using tokenkeg
* All tests should pass or be updated to pass

gitteri:
#41 completed this.

Timzkhid:
token22 support added

PR #34, opened 2025-02-28
Added token interface trait and implemented for SPL tokens

pplanel:
This PR introduces a TokenInterface trait to abstract token interactions, making it easier to support additional token standards like Token22 in the future. It includes an initial implementation for SPL tokens via the TokenKeg struct.. 
 This pull request was created for https://app.gib.work/bounties/729a6ad1-ac33-4258-ae63-e5d8180aa4ba in an attempt to solve a bounty #17 . Payment for the bounty is immediately sent to the contributor after merge.


crates/rpc/src/method/transfer_transaction.rs
```diff
@@ -66,56 +67,52 @@ pub async fn transfer_transaction(
...
-            instructions.push(create_associated_token_account(
-                &fee_payer,
-                &destination,
-                &token_mint,
-                &spl_token::id(),
-            ));
+    match token {
+        TokenType::TokenKeg(_) => {
+            instructions.push(system_instruction::transfer(&source, &destination, request.amount));
+        }
```
gitteri: This instruction is for SOL transfers. Please update this to use the token trait to avoid casing on the token types

crates/lib/src/token/types.rs
```diff
@@ -0,0 +1,166 @@
...
+        // Verify the mint owner matches the token program
+        let token_program = T::default();
+        if mint_account.owner != token_program.id() {
+            anyhow::bail!("Invalid mint owner for this token type");
+        }
+
+        Ok(Token {
+            mint: mint_data,
+            mint_address: *mint,
+            price_info: TokenPriceInfo { price: 0.0 }, // need to implement this
```
gitteri: Why are you adding price info? I don't think we need this

gitteri (changes requested):

gitteri:
hey @pplanel please take a look at this other PR for related comments. the trait should be used everywhere the spl_token crate is currently used https://github.com/solana-foundation/kora/pull/35

pplanel:
Hi @gitteri, can I update the `test_estimate_transaction_fee_with_token_creation` test case to use a real token addr?
Since the instantiation of a TokenType loads hydrate from the RPC, the test fails with a invalid token addrs.

gitteri:
> Hi @gitteri, can I update the `test_estimate_transaction_fee_with_token_creation` test case to use a real token addr? Since the instantiation of a TokenType loads hydrate from the RPC, the test fails with a invalid token addrs.

You should use a mock rpc to get fake information as needed (based on real data you can get from interacting with the real chain)

PR #42, opened 2025-03-18
Improve Docker Configuration and Configuration Management

gitteri:
# Improve Docker Configuration and Configuration Management

## Summary
This PR enhances the Kora RPC server's configuration management and Docker setup to follow best practices. It addresses several key areas:

1. Improved Docker configuration with proper port exposure and environment handling
2. Enhanced configuration file discovery with support for standard system locations
3. Removed the requirement for a private key when running the RPC server
4. Standardized port configuration across codebase

### Docker Configuration Improvements
- Updated Dockerfile to follow best practices for environment variables
- Removed hardcoded `.env` file from the Docker image for better security
- Added clear documentation in the Dockerfile for configuration options
- Created proper configuration directory structure in the container
- Standardized on port 8080 across all configuration files
- Updated docker-compose.yml to match the new configuration

### Configuration File Management
- Added a new `find_config_file` function that searches for configuration in standard locations:
  - Explicit path provided via `--config`
  - Current directory (`./kora.toml`)
  - XDG config directory (`~/.config/kora/kora.toml`)
  - System-wide config (`/etc/kora/kora.toml`)
- Updated the main application to use this new function and provide clear logging about which config file is being used

### Private Key Handling
- Made the private key optional when running the RPC server
- Updated error handling to provide clearer messages when a signer is needed but not provided
- Improved documentation around different signer options

## Testing
The changes have been tested with:
- Local development environment
- Docker container builds and runs
- Integration tests against the containerized service

### Docker Run Command
```bash
# Basic run with environment variables
docker run -e RPC_URL=https://api.devnet.solana.com -p 8080:8080 kora-rpc

# With custom config file
docker run -v $(pwd)/my-config.toml:/etc/kora/kora.toml -p 8080:8080 kora-rpc
```

### Docker Compose
```yaml
services:
  kora-rpc:
    build: .
    ports:
      - "8080:8080"
    environment:
      - RPC_URL=https://api.devnet.solana.com
      - RUST_LOG=info
    volumes:
      - ./config/kora.toml:/etc/kora/kora.toml
```

## Impact
These changes make Kora more flexible and secure by:
- Following container best practices for configuration
- Providing multiple ways to configure the application
- Improving security by not requiring sensitive information in the Docker image
- Making the system more adaptable to different deployment environments

### DIFF
--- a/crates/lib/src/args.rs
+++ b/crates/lib/src/args.rs
@@ -10,7 +10,7 @@ pub struct CommonArgs {
     pub rpc_url: String,
 
     /// Base58-encoded private key for signing
-    #[arg(long, env = "KORA_PRIVATE_KEY", required_unless_present_any = ["skip_signer", "turnkey_signer"])]
+    #[arg(long, env = "KORA_PRIVATE_KEY")]
     pub private_key: Option<String>,
 
     /// Path to kora.toml config file
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -1,11 +1,45 @@
 use serde::{Deserialize, Serialize};
-use std::{fs, path::Path};
+use std::{fs, path::{Path, PathBuf}};
 use toml;
 use utoipa::ToSchema;
 
 use solana_client::nonblocking::rpc_client::RpcClient;
 
 use crate::{error::KoraError, oracle::PriceSource, token::check_valid_tokens};
+
+/// Find the configuration file by checking multiple standard locations
+pub fn find_config_file(explicit_path: Option<&str>) -> Option<PathBuf> {
+    // 1. Check explicit path if provided
+    if let Some(path) = explicit_path {
+        let p = PathBuf::from(path);
+        if p.exists() {
+            return Some(p);
+        }
+    }
+
+    // 2. Check current directory
+    let current_dir = PathBuf::from("kora.toml");
+    if current_dir.exists() {
+        return Some(current_dir);
+    }
+
+    // 3. Check XDG config directory
+    if let Some(mut config_dir) = dirs::config_dir() {
+        config_dir.push("kora");
+        config_dir.push("kora.toml");
+        if config_dir.exists() {
+            return Some(config_dir);
+        }
+    }
+
+    // 4. Check /etc/kora/kora.toml
+    let etc_config = PathBuf::from("/etc/kora/kora.toml");
+    if etc_config.exists() {
+        return Some(etc_config);
+    }
+
+    None
+}
 
 #[derive(Debug, Deserialize)]
 pub struct Config {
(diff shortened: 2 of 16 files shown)


crates/lib/src/args.rs
```diff
@@ -10,7 +10,7 @@ pub struct CommonArgs {
     pub rpc_url: String,
 
     /// Base58-encoded private key for signing
-    #[arg(long, env = "KORA_PRIVATE_KEY", required_unless_present_any = ["skip_signer", "turnkey_signer"])]
+    #[arg(long, env = "KORA_PRIVATE_KEY")]
```
pkxro: Why do you need to remove this? ther required_unless_present makes this optional if you have external or no signer

crates/lib/src/config.rs
```diff
@@ -1,11 +1,45 @@
...
+    }
+
+    // 4. Check /etc/kora/kora.toml
+    let etc_config = PathBuf::from("/etc/kora/kora.toml");
+    if etc_config.exists() {
+        return Some(etc_config);
+    }
+
+    None
+}
```
pkxro: toml should just be in root, not sure why you need this

crates/lib/src/config.rs
```diff
@@ -1,12 +1,49 @@
...
+
 #[derive(Debug, Deserialize)]
```
pkxro: This feels like a misnomer to me -- is this for the docker file? Copying the kora.toml into the docker container should be sufficient?

crates/lib/src/oracle/fake.rs
```diff
@@ -0,0 +1,14 @@
+use super::{PriceSource, TokenPrice};
+use crate::error::KoraError;
+use reqwest::Client;
+
+/// A fake price oracle implementation that always returns a fixed price
+/// regardless of the token mint address.
+///
+/// This is useful for testing and development environments where
+/// real price data is not needed or available.
+pub async fn get_price(_client: &Client, _mint_address: &str) -> Result<TokenPrice, KoraError> {
```
pkxro: Why is this needed? Why is mockall not sufficient -- seems like an extra file that shouldn't be here

pkxro (changes requested):
Small nits

crates/lib/src/args.rs
gitteri: This was doing a thing where it was forcing the private key to be passed in as a cli arg even though it was set in an env variable. I can look to see if there's a workaround

crates/lib/src/oracle/fake.rs
gitteri: whoops yea need to remove this

gitteri: I actually do need a fake for devnet since jup doesn't work there but that can be a follow up PR

crates/lib/src/config.rs
gitteri: This is used if the app is running in docker or from the cli directly. I can remove the XDG config directory but I need the others for:
1. explicit conf path is provided
2. running from cli defaulting to current directory
3. running from docker so it's set at /etc/kora/kora.toml

PR #50, opened 2025-04-17
Enforce Base64 Transaction Encoding

gitteri:
# Enforce Base64 Transaction Encoding

## Changes
- Standardized transaction encoding to use b64 format
- Updated all tests and usages of b58 to use b64

## Purpose
This PR ensures consistent use of base64 encoding for transactions throughout the application, as it's the standard format used by Solana. This change:
- Maintains compatibility with Solana's transaction format
- Simplifies transaction handling by using a single encoding format

## Testing
- All integration tests pass
- All unit tests pass


tests/api_integration_tests.rs
```diff
@@ -38,8 +37,6 @@ async fn setup_rpc_client() -> Arc<RpcClient> {
 }
 
 async fn create_test_transaction() -> String {
-    let token_interface = TokenProgram::new(TokenType::Spl);
-    let program_id = token_interface.program_id();
```
gitteri, line 42: just clean up. these weren't used

pkxro (approved):

PR #63, opened 2025-05-01
Update README.md

Trueluvcele:
We should also include a sample example nextjs folder to test out the action(to be included in /src/app/api/action/transferTx/route.ts , or we only include the route.ts file for the action and test it out in another nextjs folder and put that demo?

Also where should the route.ts file be included in this repo?


Trueluvcele:
Ok

PR #70, opened 2025-05-30
feat: Added in privy support

ControlCplusControlV:
Added in a privy signer which uses, adding a draft, but let me know what the best way to test this. Happy to hand over the other rust-sdk Repo as well


pkxro:
closed for #80

PR #99, opened 2025-07-16
Feature/improved fee payer protection

dev-jodee:


crates/lib/src/config.rs
```diff
@@ -13,6 +13,25 @@ pub struct Config {
...
+    pub allow_assign: bool,
+}
+
+impl Default for FeePayerPolicy {
+    fn default() -> Self {
+        Self {
+            allow_sol_transfers: true,
+            allow_spl_transfers: true,
+            allow_token2022_transfers: true,
+            allow_assign: true,
```
amilz, line 30: I understand that we'd want to have this backward compatible, but if we don't think there's any real application why any node operator would want to do this--I think it'd make sense to default false and include a warning of the change on install (in case an existing operator upgrades).

crates/lib/src/transaction/validator.rs
```diff
@@ -183,18 +185,79 @@ impl TransactionValidator {
...
+                if let Ok(sys_ix) =
+                    bincode::deserialize::<system_instruction::SystemInstruction>(&ix.data)
+                {
+                    match sys_ix {
+                        system_instruction::SystemInstruction::Transfer { .. }
+                        | system_instruction::SystemInstruction::TransferWithSeed { .. } => {
+                            if !self.fee_payer_policy.allow_sol_transfers {
+                                // For transfer instruction, first account is source
+                                return account_keys[ix.accounts[0] as usize]
+                                    == self.fee_payer_pubkey;
```
amilz: I think `TransferWithSeed` signer is actually position 1. I think position 0 here is the derived account. 

Also need to add a test case with `test_fee_payer_policy_assign`

```rust
    /// Transfer lamports from a derived address
    ///
    /// # Account references
    ///   0. `[WRITE]` Funding account
    ///   1. `[SIGNER]` Base for funding account
    ///   2. `[WRITE]` Recipient account
    TransferWithSeed {
        /// Amount to transfer
        lamports: u64,

        /// Seed to use to derive the funding account address
        from_seed: String,

        /// Owner to use to derive the funding account address
        from_owner: Pubkey,
    },
```

crates/lib/src/transaction/validator.rs
```diff
@@ -183,18 +185,79 @@ impl TransactionValidator {
...
+                            }
+                        }
+                        system_instruction::SystemInstruction::Assign { .. } => {
```
amilz, line 203: TIL -- there's an `AssignWithSeed` that we might want to include.

crates/lib/src/transaction/validator.rs
```diff
@@ -739,6 +809,268 @@ mod tests {
         assert!(validated_amount > amount, "Interest should be added to the amount");
     }
 
+    #[test]
+    fn test_fee_payer_policy_sol_transfers() {
```
amilz, line 810: add check for `transfer_checked`

amilz (changes requested):
This is awesome. See comments below. 
I feel like it is probably worth doing this for all system/token program instructions tbh--particularly burn/close/approve.

crates/lib/src/config.rs
dev-jodee: Like we talked about, Assign is often use when creating accounts, which would most likely mean that the payer of the rent would be Kora itself (since the whole point of use a gasless relayer is not needing to have SOL), so the node operator would most likely allow by default the creation of new accounts (since a lot of the transactions on solana need it).

But it does come with its own risks, which is why I think we should still have it as a flag, but by default true, since account creation is extremely frequent in the Solana ecosystem.

crates/lib/src/transaction/validator.rs
dev-jodee: absolutely right thanks !

crates/lib/src/transaction/validator.rs
dev-jodee: you mean transfer with seed ? Transfer checked doesn't exist on sol transfers

pkxro (approved):

PR #101, opened 2025-07-17
fix: privy expecting a serialized transaction rather than serialized m…

dev-jodee:
…essage. Fix make install to only install the crates with bins.

### DIFF
--- a/crates/lib/src/signer/signer.rs
+++ b/crates/lib/src/signer/signer.rs
@@ -2,7 +2,7 @@ use super::{solana_signer::SolanaMemorySigner, vault_signer::VaultSigner};
 use crate::error::KoraError;
 use kora_privy::PrivySigner;
 use kora_turnkey::TurnkeySigner;
-use solana_sdk::signature::Signature as SolanaSignature;
+use solana_sdk::{signature::Signature as SolanaSignature, transaction::Transaction};
 use std::error::Error;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


crates/lib/src/signer/signer.rs
```diff
@@ -52,32 +52,37 @@ impl KoraSigner {
...
-            KoraSigner::Vault(signer) => signer.sign_solana(message).await,
+            // Some signers expect the serialized message, others expect the message bytes
+            KoraSigner::Memory(signer) => signer.sign_solana(transaction).await,
+            KoraSigner::Vault(signer) => signer.sign_solana(transaction).await,
             KoraSigner::Turnkey(signer) => {
-                signer.sign_solana(message).await.map_err(KoraError::from)
+                signer.sign_solana(&transaction.message_data()).await.map_err(KoraError::from)
             }
-            KoraSigner::Privy(signer) => signer.sign_solana(message).await.map_err(KoraError::from),
+            KoraSigner::Privy(signer) => signer
```
amilz: a nit, but for consistency wdyt about having *Privy* and *Turnkey* `sign_solana` also take in the `transaction` for consistency and then `serialize` or extract `message_data` within the `sign_solana` call.

amilz (approved):
Thanks! Tested all methods w/ Privy & Memory signers. 1 minor suggestion but I think it's fine as-is.

pkxro (approved):

pkxro (approved):

PR #103, opened 2025-07-17
feat: Improve signer parsing with support for byte arrays and .json files

dev-jodee:


tests/testing-utils/src/lib.rs
amilz, line 1: Between`FEE_PAYER_DEFAULT`, `SENDER_SIGNER_DEFAULT`, and `TEST_USDC_MINT_KEYPAIR`, we have full coverage here--so I think we're good...but long term, it's probably worth adding a utils unit tests for

amilz (approved):
lgtm--thanks! `test result: ok. 13 passed;`

pkxro (approved):

PR #104, opened 2025-07-17
fix: local TS SDK Tests 

amilz:
- updated setup to use local keys with mint creation for consistent testing across environments
- fixed ESM issues
- updated and cleaned up dependencies in package.json
- added new test script with local test validator

Should be able to 

```bash
make run
```

Seperate terminal 

```bash
cd sdks/ts
pnpm install
pnpm test:integration
```


dev-jodee (approved):
lgtm

PR #108, opened 2025-07-21
feat: Update dependencies and refactor transaction handling to support Vers…

dev-jodee:
…ionTransaction

- Added new Solana dependencies: `solana-commitment-config`, `solana-message`, and `solana-system-interface`.
- Refactored transaction handling to use `VersionedTransaction` and `VersionedMessage` across various modules.
- Updated `Cargo.toml` files to reflect new dependencies and features.
- Enhanced transaction validation and signing processes to support versioned transactions.
- Improved test cases to cover new transaction structures and lookup table validations.


crates/lib/src/transaction/transaction.rs
```diff
@@ -1,18 +1,38 @@
...
+        signatures: vec![Signature::default(); num_required_signatures],
+        message,
+    }
+}
+
+pub async fn get_estimate_fee(
+    rpc_client: &RpcClient,
+    message: &VersionedMessage,
+) -> Result<u64, KoraError> {
+    match message {
```
pkxro, line 30: Why do we ned to match here if we're calling the same underlying rpc call?

pkxro (approved):
Looks generally good, just one small nit

amilz:
PRO-29

crates/lib/src/transaction/transaction.rs
dev-jodee: Yeah it's because VersionedMessage is 

#[derive(Debug, PartialEq, Eq, Clone)]
pub enum VersionedMessage {
    Legacy(LegacyMessage),
    V0(v0::Message),
}

and v0::Message and LegacyMessage are the struct impl SerializableMessage (which is what the RPC calls expect as parameter), so we still need to do the match :/ I'm also annoyed by it

PR #113, opened 2025-07-23
chore: remove Jupiter dependency and update price source handling

dev-jodee:
- Removed Jupiter dependency from Cargo.toml and related workspace configurations.
- Updated price source handling in the codebase to accommodate new structure with optional API URL.
- Enhanced test cases to validate new price source configurations and ensure backward compatibility.


amilz:
PRO-60

crates/lib/src/config.rs
```diff
@@ -104,7 +160,11 @@ mod tests {
         assert_eq!(config.validation.allowed_tokens, vec!["token1", "token2"]);
         assert_eq!(config.validation.allowed_spl_paid_tokens, vec!["token3"]);
         assert_eq!(config.validation.disallowed_accounts, vec!["account1"]);
-        assert_eq!(config.validation.price_source, PriceSource::Jupiter);
+        if let PriceSource::Jupiter { api_url: _ } = config.validation.price_source {
+            // Test passes - Jupiter source was correctly parsed
+        } else {
+            panic!("Expected Jupiter price source");
+        }
```
amilz: maybe an assertion like this?
also don't you need to update your `config_content` based on new jupiter enum?

```suggestion
        assert!(matches!(config.validation.price_source, PriceSource::Jupiter { .. }));
```

crates/lib/src/oracle/oracle.rs
amilz, line 1: I think we might actually need an optional API key that passes to the header. 

---- https://dev.jup.ag/docs/api-setup

For paid tiers with API Keys, use api.jup.ag
For free tier, use lite-api.jup.ag (NO API Key required)
API Key

Simply add the API Key in the x-api-key field in the Headers.

```
headers: {
    'Content-Type': 'application/json',
    'x-api-key': '' // enter api key here
},
```

crates/lib/src/oracle/jupiter.rs
```diff
@@ -55,18 +69,20 @@ impl PriceOracle for JupiterPriceOracle {
...
-            .data
             .get(mint_address)
             .ok_or_else(|| KoraError::RpcError("No price data from Jupiter".to_string()))?;
 
-        // Convert price from string to f64
-        let price = price_data
-            .price
-            .parse::<f64>()
-            .map_err(|e| KoraError::RpcError(format!("Failed to parse price: {e}")))?;
+        let price = price_data.usd_price / sol_price.usd_price;
```
amilz, line 154: Is this change intentional?

dev-jodee: Yes, so before the v2 API in the url parameters you specify i.e. JUP vsToken SOL and would get the SOL price 

`"{}/price?ids={}&vsToken=So11111111111111111111111111111111111111112",`

But with v3, you do not have vsToken, so I get the prices for both SOL and XYZ token, the prices are returned in USD so to get the price in SOL we divide both :)

https://dev.jup.ag/docs/price-api/v3

amilz: thx--good catch.

crates/lib/src/oracle/jupiter.rs
```diff
@@ -76,40 +138,100 @@ mod tests {
...
+        // Test env var fallback when config is None
+        let oracle_env = JupiterPriceOracle::new(None);
+        assert_eq!(oracle_env.api_key, Some("321".to_string()));
+
+        std::env::remove_var("JUPITER_API_KEY");
+
+        // Test default fallback when both config and env are None
+        let oracle_default = JupiterPriceOracle::new(None);
+        assert_eq!(oracle_default.api_key, None);
+    }
```
amilz: WDYT about adding a Jup integration test in here to make sure our API calls are working as expected?
I think you could pick some stablecoin like USDC and expect some range of acceptable results?

crates/lib/src/oracle/jupiter.rs
```diff
@@ -1,30 +1,52 @@
...
+    usd_price: f64,
+    #[serde(rename = "blockId")]
+    #[allow(dead_code)]
+    block_id: u64,
+    #[allow(dead_code)]
+    decimals: u8,
+    #[serde(rename = "priceChange24h")]
     #[allow(dead_code)]
-    time_taken: f64,
+    price_change_24h: f64,
```
amilz: I got some errors trying to test this on USDC. Looks like maybe the response doesn't *always* include this field. I changed to `price_change_24h: Option<f64>` and it resolved the err.

Response:

```json
{
  "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v": {
    "usdPrice": 0.999902412255114,
    "blockId": 355169121,
    "decimals": 6
  }
}
```

amilz (commented):
Couple of thoughts: 
1. WDYT about adding the jup api key to *CommonArgs* in `crates/lib/src/args.rs` so we can use CLI consistent w/ how we do signers...and fallback to env
2. I think we need Jup integration tests (and may need to update our response struct based on how those work) -- i can add this

crates/lib/src/oracle/jupiter.rs
dev-jodee: yes agreed

crates/lib/src/oracle/jupiter.rs
dev-jodee: jup

Yeah you're right, they have it in the example, but seems like it's not in the parameters they describe :')

crates/lib/src/oracle/jupiter.rs
```diff
@@ -1,30 +1,52 @@
...
+    pro_api_url: String,
+    lite_api_url: String,
+    api_key: Option<String>,
 }
 
-pub struct JupiterPriceOracle;
+impl JupiterPriceOracle {
+    pub fn new(api_key: Option<String>) -> Self {
+        // Check environment variable for API key
+        let api_key = api_key.or_else(|| std::env::var("JUPITER_API_KEY").ok());
```
pkxro: Can we do this on load and not here?

pkxro (approved):

PR #114, opened 2025-07-23
feat: add pricing model configuration for transaction validation (PRO-56)

dev-jodee:
**For now would still only support 1 price at a time, possibility to add multi prices later on**

Why?

Let the node operators decide on different pricing models, instead of only relying on margin of estimated lamport fee.

What?

- Introduced new pricing models: margin, fixed, and free in `kora.toml`.
- Updated README to document pricing model configurations.
- Refactored transaction signing logic to utilize the new pricing model.
- Enhanced tests to cover new pricing model functionality and validation.

### DIFF
--- a/sdks/.changeset/quick-animals-attack.md
+++ b/sdks/.changeset/quick-animals-attack.md
@@ -1,0 +1,5 @@
+---
+"@kora/sdk": minor
+---
+
+update TS SDK types for changes to Kora's price model config (add PriceConfig to Validator Config) and remove margin from transaction cost estimate param


crates/lib/src/transaction/fees.rs
```diff
@@ -19,6 +19,55 @@ pub struct TokenPriceInfo {
     pub price: f64,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(tag = "type", rename_all = "lowercase")]
+pub enum PriceModel {
+    Margin { margin: f64 },
+    Fixed { amount: u64, token: String },
+    Free,
```
pkxro, line 27: Why do we need a free enum if margin of 0 is free as the default?

dev-jodee: Margin of 0 doesn't mean it's free, it means the node operator doesn't add a margin on top of the txn fee (but the fee is still expected to be paid). Free = not even the fee

PR #116, opened 2025-07-23
feat: enhance Kora configuration and RPC method handling (PRO-53)

dev-jodee:
Why?

Let the node operator decide which rpc function they want to support.

What?

* Updated Kora configuration to include enabled methods for RPC.
* Introduced a new struct `EnabledMethods` to manage method availability.
* Modified RPC server to register methods based on enabled configuration.
* Added integration tests for method availability based on configuration.
* Updated documentation for local server setup and testing configurations.


CLAUDE.md
```diff
@@ -71,10 +71,11 @@ Integration tests require a local validator and test account setup:
    solana-test-validator --reset --quiet
    ```
 
-2. **Start local Kora Server:**
+2. **Start local Kora Server with test configuration:**
     ```bash
-    make run
+    cargo run -p kora-rpc --bin kora-rpc -- --private-key ./tests/testing-utils/local-keys/fee-payer-local.json --config tests/kora-test.toml --rpc-url http://127.0.0.1:8899
```
amilz, line 76: we could make a `make run:test` or something just to clean this up a little. and make sure to update readme to match.

crates/lib/src/config.rs
```diff
@@ -45,10 +45,40 @@ pub struct ValidationConfig {
     pub fee_payer_policy: FeePayerPolicy,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct EnabledMethods {
```
amilz, line 49: WDYT about expsosing this in `ValidationConfig` so the user can fetch via `getConfig` ?

amilz (commented):
LGTM--all tests passing. couple of nit comments.

CLAUDE.md
dev-jodee: Yeah, I did in the auth pr, so i'll resolve the comment here :)

crates/lib/src/config.rs
dev-jodee: yes for sure, my mistake, it only shows validation config not the kora config

PR #119, opened 2025-07-24
feat: (PRO-69) Implement API Key and HMAC Authentication for Kora RPC

dev-jodee:
Why?

So node operators can control who has access to their node and also control integrity of the requests.

What?

- Added support for API Key and HMAC authentication methods in the Kora RPC server.
- Updated configuration to include optional `api_key` and `hmac_secret`.
- Enhanced integration tests to cover authentication scenarios.
- Updated documentation to reflect new authentication methods and usage examples.
- Refactored Makefile and CI to streamline integration test execution.

### DIFF
--- a/crates/rpc/src/auth.rs
+++ b/crates/rpc/src/auth.rs
@@ -1,0 +1,436 @@
+use futures_util::TryStreamExt;
+use hmac::{Hmac, Mac};
+use http::{Request, Response, StatusCode};
+use jsonrpsee::server::logger::Body;
+use kora_lib::constant::{X_API_KEY, X_HMAC_SIGNATURE, X_TIMESTAMP};
+use sha2::Sha256;
+
+const MAX_TIMESTAMP_AGE: i64 = 300; // 5 minutes in seconds (could make this configurable?)
+
+pub async fn extract_parts_and_body_bytes(
+    request: Request<Body>,
+) -> (http::request::Parts, Vec<u8>) {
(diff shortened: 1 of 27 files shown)


amilz (commented):
- Args: wdyt about adding `--api-key` and `--hmac-secret` flags to `CommonArgs` in `crates/lib/src/args.rs` so folks can deploy using env secrets?
- wdyt about adding a `make generate-key` command that calls `openssl rand -hex 32` ?
- updated [docs](https://github.com/solana-foundation/kora/pull/124) based on how i'm hoping to structure all of this.
- SDK: 
    - we should add headers to our configuration 
    - we should add helper method for `createHmacSignature`

crates/rpc/src/auth.rs
```diff
@@ -0,0 +1,436 @@
+use futures_util::TryStreamExt;
+use hmac::{Hmac, Mac};
+use http::{Request, Response, StatusCode};
+use jsonrpsee::server::logger::Body;
+use kora_lib::constant::{X_API_KEY, X_HMAC_SIGNATURE, X_TIMESTAMP};
+use sha2::Sha256;
+
+const MAX_TIMESTAMP_AGE: i64 = 300; // 5 minutes in seconds (could make this configurable?)
```
pkxro, line 8: Would be good to make this configruable in the next round as part of the config

pkxro (approved):
Why do we need two types here? We should just pick one and stick with it (probably just api key)

PR #123, opened 2025-07-24
feat: (PRO-70) implement compute budget handling for transaction fees…

dev-jodee:
… (calculate priority fee instead of estimating)

Why?

Better than estimating, since we can just find the compute budget instructions

What?

- Added support for compute budget instructions in transaction fee estimation.
- Introduced `VersionedTransactionResolved` for resolving addresses in V0 transactions. This was added because of not wanting to load the lookup table addresses multiple time per transaction. When / If we ever implement a cache (i.e. REDIS), this could be a good place to store those lookup table.
- Updated transaction signing and validation to utilize resolved addresses.
- Enhanced documentation on lookup table resolution and compute budget integration.
- Added integration tests for compute budget scenarios.


crates/lib/src/transaction/fees.rs
```diff
@@ -1,28 +1,82 @@
...
+}
+
+impl ComputeUnitInfo {
+    pub fn calculate_priority_fee(&self) -> u64 {
+        let limit = self.compute_unit_limit.unwrap_or(BASE_COMPUTE_UNIT_LIMIT);
+        let price = self.compute_unit_price.unwrap_or(BASE_COMPUTE_UNIT_PRICE);
+
+        // Priority fee = compute_unit_limit * compute_unit_price
+        // Note: compute_unit_price is in microlamports, so we divide by 1_000_000 to get lamports
+        (limit as u64).saturating_mul(price) / 1_000_000
```
amilz: maybe add to constants `MICRO_LAMPORTS_LAMPORT`

crates/lib/src/transaction/transaction.rs
amilz, line 1: would it make sense here to add some unit tests to make sure `resolve_addresses` is working as intended?

tests/compute_budget_integration_tests.rs
```diff
@@ -0,0 +1,101 @@
...
+        )
+        .await
+        .expect("Failed to estimate transaction fee");
+
+    assert!(response.get("fee_in_lamports").is_some(), "Response should have result field");
+    let fee = response["fee_in_lamports"].as_u64().expect("Fee should be a number");
+
+    // Fee should include priority fee from compute budget instructions
+    // Priority fee calculation: 300_000 * 50_000 / 1_000_000 = 15_000 lamports
+    // Plus base transaction fee (20 000 for this transaction) = 35_000 lamports total
```
amilz: shouldn't base fee be 2 x 5,000? Same for next test as well.

amilz (commented):
Nice work. All tests running for me.

crates/lib/src/transaction/fees.rs
dev-jodee: yeah I can

tests/compute_budget_integration_tests.rs
dev-jodee: getFeeforMessage included priority fees in the end so changed everything

pkxro (approved):

PR #124, opened 2025-07-25
docs: Add detailed authentication guide and update operator README (PRO-78)

amilz:
Creating an Auth doc more and changing op readme to more general overview consistent with other docs being worked on.


dev-jodee (approved):
nice

PR #130, opened 2025-08-04
feat: (PRO-82) add configuration validation enhancements and CLI options

dev-jodee:
- Introduced `validate_config` flag in CLI for optional configuration validation.
- Added `validate_with_result` method in `Config` for detailed validation results.
- Updated Makefile to streamline linting and formatting commands.
- Refactored validation logic to provide warnings and errors for configuration issues.
- Enhanced tests for validation scenarios, ensuring comprehensive coverage of edge cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: check fmt lint lint-fix test build run clean all regen-tk fix-all generate-ts-client setup-test-env test-integration test-integration-coverage coverage
+.PHONY: check lint test build run clean all install generate-key setup-test-env test-integration test-integration-coverage test-all test-ts coverage coverage-clean build-bin build-lib build-rpc build-tk run-presigned openapi gen-ts-client
 
 # Common configuration
 TEST_PORT := 8080
@@ -25,17 +25,11 @@ install:
 check:
 	cargo fmt --all -- --check
 
-# Format code
-fmt:
-	cargo fmt --all
-
-# Run clippy
+# Run all fixes and checks
 lint:
-	cargo clippy
-
-# Run clippy with auto-fix
-lint-fix:
-	cargo clippy --fix --allow-dirty
+	cargo clippy --fix --allow-dirty -- -D warnings
+	cargo fmt --all
+	cargo fmt --all -- --check
 	
 # Run tests
 test:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -33,10 +27,11 @@ setup-test-env:
 	cargo run -p tests --bin setup-test-env
 
 # Run integration tests
-test-integration:
-	cargo run -p tests --bin setup-test-env
+test-integration: setup-test-env
 	cargo test --test integration
 
+test-all: test test-integration
```
amilz: wdyt about adding client tests to this command?

crates/lib/src/config.rs
```diff
@@ -164,6 +206,130 @@ impl Config {
...
+            warnings.push("Max signatures is 0 - this will block all transactions".to_string());
+        }
+
+        // Validate price source (warn if Mock)
+        if matches!(self.validation.price_source, PriceSource::Mock) {
+            warnings.push("Using Mock price source - not suitable for production".to_string());
+        }
+
+        // Validate allowed programs (warn if empty or missing system/token programs)
+        if self.validation.allowed_programs.is_empty() {
```
amilz, line 245: wdyt about adding a `check_valid_program` to the program validation...valid address && owner = `BPFLoaderUpgradeab1e11111111111111111111111`

crates/lib/src/config.rs
amilz, line 1: wdyt about fee_payer_policy warnings?

amilz (approved):
lgtm. couple suggestions

crates/lib/src/config.rs
dev-jodee: Yeah I thought about it, but I'm scared that validation using RPC will be kind of problematic, we could validate tokens, programs, etc. and check if signer, owner, etc. but RPCs aren't always stable and it could potentially spam the RPC. For now I prefer non rpc check but maybe if we get asked we could go with the full validation

crates/lib/src/config.rs
dev-jodee: Yeah thought about it but honestly not 100% sure what the warning could be, like assign is disabled should we just flag it or explain what it disables, and if so then the warnings might become very annoying? What do you think

amilz: fair. maybe at some point we could add a `verbose` flag or something if folks ask for more info. i think fine as-is for now.

crates/lib/src/config.rs
amilz: sgtm

PR #135, opened 2025-08-05
feat(metrics): add metrics collection and monitoring for Kora RPC (PRO-61)

dev-jodee:
…O-61)

- Introduced a new `kora-metrics` crate for Prometheus metrics.
- Implemented HTTP metrics collection with `HttpMetricsLayer` and `MetricsHandlerLayer`.
- Updated Docker and Docker Compose configurations for metrics services (Prometheus and Grafana).
- Enhanced `Makefile` with new targets for running the monitoring stack.
- Added Grafana dashboard and Prometheus configuration for Kora metrics.
- Updated `Cargo.toml` and `.env.example` to include new dependencies and environment variables for metrics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2976,6 +2976,8 @@ dependencies = [
  "clap",
  "deadpool-redis",
  "dirs",
+ "futures-util",
+ "http 0.2.12",
  "jsonrpsee",
  "jup-ag",
  "kora-privy",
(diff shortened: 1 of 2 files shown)


amilz (commented):
Nice addition.
I tested w/o docker and just doing some curl commands to `/metrics` and I ran through docker as well and got the UI up and running. 

A couple of thoughts for discussion:

#### 1. Runtime Configuration for Metrics
I'm not sure how much overhead having this `/metrics` endpoint adds, but I think it could make sense to add runtime control to match our existing patterns:

**Option A: CLI Flag**
```bash
kora-rpc --disable-metrics  # Simple on/off
```

**Option B: Config Section**
```toml
[metrics]
enabled = true
metrics-endpoint = "/metrics"  # Allow custom path
port = 8080           # Or separate port like 9090
# maybe future enabled_metrics = []
```

We already have `--metrics-endpoint` flag in our cli--I don't think it actually does anything...I think we should try to hook up one of the options above or get rid of this cli arg.


#### 2. Docker Architecture

Current setup`docker-compose.metrics.yml` bundles Kora + monitoring stack
- This can create conflicts if users already run Kora locally (happened to me).
- Couples application deployment with monitoring (which may not always be desired)

WDYT about seperating concerns where this file just starts a metrics service? and we could have a kora-rpc one too?  Alternative If keeping bundled, maybe rename to `docker-compose.full-stack.yml` to clarify it includes everything.

amilz (approved):

PR #146, opened 2025-08-07
refactor: Main PR representing refactoring & code clean up & organiza…

dev-jodee:
…tion before release

- Introduced `FeeConfigUtil` for improved transaction fee estimation and fee payer outflow calculations.
- Reorganized the code in different folder that are linked (validators, fees, etc.)
- Cleaned up mod.rs's so they don't contain code
- Created testing utils for unit testing to reuse functions
- Reorganized unit tests to be in the same file as the code they are testing, removed external unit tests files

### DIFF
--- a/docs/operators/AUTHENTICATION.md
+++ b/docs/operators/AUTHENTICATION.md
@@ -11,6 +11,8 @@ Authentication is optional but **strongly recommended** for production deploymen
 | **HMAC** | High | Public APIs, untrusted networks | Medium |
 | **Both** | Maximum | High-security environments | Medium |
 
+Kora Authentication is configured in the `kora.toml` file in the `[kora.auth]` section.
+
 In this document:
 
 - [API Key Authentication](#api-key-authentication)
@@ -36,8 +38,7 @@ openssl rand -hex 32
 - Add an `api_key` to your `kora.toml`:
 
 ```toml
-[kora]
-rate_limit = 100
+[kora.auth]
 api_key = "kora_live_sk_1234567890abcdef"  # Use a strong, unique key
 ```
 
@@ -91,8 +92,7 @@ Instead of sending an API key with every request, HMAC creates a unique cryptogr
 - Add a global `hmac_secret` to your `kora.toml` (minimum 32 characters--you can generate one with `openssl rand -hex 32` or similar):
 
 ```toml
-[kora]
-rate_limit = 100
+[kora.auth]
 hmac_secret = "kora_hmac_your-strong-hmac-secret-key"
 ```
 
@@ -156,8 +156,7 @@ console.log(config);
 You can enable both authentication methods simultaneously for maximum security:
 
 ```toml
-[kora]
-rate_limit = 100
+[kora.auth]
 api_key = "kora_live_sk_1234567890abcdef"
 hmac_secret = "kora_hmac_your-strong-hmac-secret-key"
 ```
--- a/docs/operators/CONFIGURATION.md
+++ b/docs/operators/CONFIGURATION.md
@@ -18,6 +18,7 @@ Your configuration file should be placed in your deployment directory or specifi
 The `kora.toml` file is organized into sections, each with its own set of options. This guide walks through each section and explains the options available:
 
 - [Kora Core Policies](#kora-core-policies) - Core server settings
+- [Kora Authentication](#kora-authentication) - Authentication settings
 - [Kora Enabled Methods](#kora-enabled-methods) - Kora RPC methods to enable
 - [Validation Policies](#validation-policies) - Transaction validation and security
 - [Fee Payer Policy](#fee-payer-policy) - Restrictions on fee payer wallet
@@ -29,6 +30,9 @@ Sample `kora.toml` file sections:
 ```toml
 [kora]
 # Core server settings
+
+[kora.auth]
+# Authentication settings
 
 [kora.enabled_methods]
 # Kora RPC methods to enable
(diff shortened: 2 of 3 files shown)


dev-jodee:
Other refactoring will be PRs against this main refactoring branch

What's left

- [ ] Token folder refactoring to streamline and unify all token / mint related checks, instructions, validations, etc. (lib/token, validator.rs, account_validator.rs)
- [ ] Streamline unit test utilities (mocked RPC, mocked accounts like Mints, etc.) and change all the unit tests to use these utilities (config_validator.rs and other files)
- [ ] CLI should be the only package that we publish and that is visible, CLI will be used to start the rpc (kora rpc, kora config, etc.)
- [ ] CLI shouldn't have any instructions like sign, sign if paid, etc.
- [ ] Move Privy and Turnkey in lib as well as the rpc crate under lib
- [ ] Transaction.rs should split the functions to their related utilities (like estimate fee should be in fees, etc.)
- [ ] Clean up and streamline integration tests to use shared utilities
- [ ] Code Coverage using CodeCov

crates/lib/src/rpc_server/args.rs
amilz, line 1: since we are cleaning things up a bit--wdyt about creating structs for different "group elements" of the CLI (e.g., auth, privy, etc.)? I think this will help make this a little more manageable over time. Something like:

```rust
/// RPC server arguments
#[derive(Debug, Parser)]
pub struct RpcArgs {
    /// HTTP port to listen on for RPC requests
    #[arg(short = 'p', long, default_value = "8080")]
    pub port: u16,

    /// Output format for logs (standard or json)
    #[arg(long, default_value = "standard")]
    pub logging_format: LoggingFormat,

    /// Prometheus metrics endpoint URL for monitoring
    #[arg(long)]
    pub metrics_endpoint: Option<String>,

    /// Authentication arguments
    #[command(flatten)]
    pub auth: AuthArgs,

    /// Private key for transaction signing (base58, u8 array, or path to JSON keypair file).
    /// Required unless `skip_signer`, `turnkey_signer`, `vault_signer`, or `privy_signer` is set
    #[arg(long, env = "KORA_PRIVATE_KEY", required_unless_present_any = ["skip_signer", "turnkey_signer", "vault_signer", "privy_signer"], help_heading = "Signing Options")]
    pub private_key: Option<String>,

    /// Skip signer initialization (useful for testing or operations that don't require signing)
    #[arg(long = "no-load-signer", help_heading = "Signing Options")]
    pub skip_signer: bool,

    /// Turnkey signer arguments
    #[command(flatten)]
    pub turnkey: TurnkeyArgs,

    /// Privy signer arguments
    #[command(flatten)]
    pub privy: PrivyArgs,

    /// Vault signer arguments
    #[command(flatten)]
    pub vault: VaultArgs,
}
```

crates/lib/src/token/spl_token_2022.rs
```diff
@@ -0,0 +1,1228 @@
...
+                return Err("Token is cpi guarded".into());
+            }
+
+            let transfer_fee = account.get_transfer_fee();
+            if transfer_fee.is_some() {
+                let basis_points = 100; // 1%
+                let fee = (amount as u128 * basis_points as u128 / 10000) as u64;
+
+                // Cap at 10,000 lamports maximum fee
+                let max_fee = 10_000;
```
amilz: wdyt about moving these to our constants?

crates/lib/src/config.rs
amilz, line 1: WDYT about doing an auth config? 

```rust
#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
pub struct KoraConfig {
    pub rate_limit: u64,
    #[serde(default)]
    pub enabled_methods: EnabledMethods,
    pub auth_config: Option<AuthConfig>,
}
```

crates/lib/src/signer/mod.rs
amilz, line 1: - Only b/c we're putting privy & turnkey in a `signer/` subdir, it might make sense to do the same with `solana_signer` and `vault_signer`--just for organizational consistency
- I think we can probably combine `signer/utils.rs` and `signer/keypair_util.rs` as both are related to keypair

amilz (approved):
Refactor LGTM and all tests are passing locally--
I dropped a couple of minor organizational suggestions for consideration.
- Let's drop unnecessary CI: 
  - `.github/workflows/kora-privy-publish.yml`, 
  - `.github/workflows/kora-rpc-publish.yml`, 
  - `.github/workflows/kora-turnkey-publish.yml`
  - `.github/workflows/kora-lib-publish.yml` (I don't think we need to publish this? wdyt)

crates/cli/src/main.rs
amilz, line 1: I think we lost our env loading and logger in the migration. Which means the env vars aren't working and if you use flags instead the console doesn't give any indication the server has started.

### dotenv

Add: 

```rust
 dotenv::dotenv().ok();
```

### Logger

Add: 

```rust
      // after  Some(Commands::Rpc { rpc_args }) => {
            setup_logging(&rpc_args.logging_format);
```

and at end of file: 

```rust
fn setup_logging(format: &LoggingFormat) {
    let env_filter = std::env::var("RUST_LOG")
        .unwrap_or_else(|_| "info,sqlx=error,sea_orm_migration=error,jsonrpsee_server=warn".into());

    let subscriber = tracing_subscriber::fmt().with_env_filter(env_filter);
    match format {
        LoggingFormat::Standard => subscriber.init(),
        LoggingFormat::Json => subscriber.json().init(),
    }
}
```

Need to add tracing to `Cargo.toml`

crates/cli/Cargo.toml
```diff
@@ -6,13 +6,18 @@ description = "CLI for Kora gasless relayer"
...
+
 [dependencies]
 kora-lib = { path = "../lib", version = "1.0.2" }
 clap = { workspace = true }
 tokio = { workspace = true }
 env_logger = { workspace = true }
 solana-client = { workspace = true }
 dotenv = { workspace = true }
+serde_json = { workspace = true }
+utoipa = { workspace = true, optional = true }
```
amilz, line 23: Need to add tracing to set up logger:

```suggestion
utoipa = { workspace = true, optional = true }
tracing = { workspace = true }
tracing-subscriber = { workspace = true }
```

amilz (changes requested):

amilz (approved):

PR #156, opened 2025-08-12
fix: (PRO-155) resolve mint account index for TransferChecked instructions

amilz:
Fixed incorrect mint account resolution in token payment validation that was causing "An account's data contents was invalid" error when running TypeScript integration test. Issue was not present on Rust integration test. This was happening b/c Rust test uses `transfer` and TS test uses `transfer_checked` (different account indices).

The issue occurred when processing TransferChecked instructions in transactions with multiple instructions (e.g., ATA creation + transfer). The mint index returned by decode_transfer_instruction is relative to the instruction's accounts, but was being used directly as an index  into the transaction's global account list.

This caused the system to attempt unpacking the wrong account (e.g. the destination ATA) as mint data, resulting in validation failures.

Changes:
- Use ix.accounts[mint_index] to properly resolve mint account index
- Add bounds checking for instruction account access

Resolves the TS integration test error that was preventing proper was was using TransferChecked for token payments.


dev-jodee (approved):
thank you

PR #173, opened 2025-08-15
feat(cli): add initialize-atas command for payment token ATAs & custo…

dev-jodee:
…m payment address for Paymaster

- Introduced a new subcommand `initialize-atas` under the RPC command to initialize associated token accounts (ATAs) for allowed payment tokens.
- Updated CLI structure to support the new command.
- Enhanced KoraConfig to include an optional payment address.
- Implemented logic to validate and create ATAs directly from the command line.
- Added tests to verify the functionality of the new command.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -3,6 +3,7 @@ mod args;
 use args::GlobalArgs;
 use clap::{Parser, Subcommand};
 use kora_lib::{
+    admin::token_util::initialize_paymaster_atas,
     error::KoraError,
     log::LoggingFormat,
     rpc::get_rpc_client,
@@ -25,10 +26,10 @@ enum Commands {
         #[command(subcommand)]
         config_command: ConfigCommands,
     },
-    /// Start the RPC server
+    /// RPC server operations
     Rpc {
-        #[command(flatten)]
-        rpc_args: Box<RpcArgs>,
+        #[command(subcommand)]
+        rpc_command: RpcCommands,
     },
     /// Generate OpenAPI documentation
     #[cfg(feature = "docs")]
@@ -45,6 +46,20 @@ enum ConfigCommands {
     Validate,
     /// Validate configuration file with RPC validation (slower but more thorough)
     ValidateWithRpc,
+}
+
+#[derive(Subcommand)]
+enum RpcCommands {
+    /// Start the RPC server
+    Start {
+        #[command(flatten)]
+        rpc_args: Box<RpcArgs>,
+    },
+    /// Initialize ATAs for all allowed payment tokens
+    InitializeAtas {
+        #[command(flatten)]
+        rpc_args: Box<RpcArgs>,
+    },
 }
 
 #[derive(Parser)]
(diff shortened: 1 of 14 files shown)


crates/lib/src/admin/token_util.rs
```diff
@@ -0,0 +1,129 @@
...
+    let blockhash = rpc_client
+        .get_latest_blockhash()
+        .await
+        .map_err(|e| KoraError::RpcError(format!("Failed to get blockhash: {e}")))?;
+
+    let message = VersionedMessage::Legacy(Message::new_with_blockhash(
+        &instructions,
+        Some(&signer.solana_pubkey()),
+        &blockhash,
+    ));
```
amilz: 1. We should at at a compute ix -- just for best practices.
2. Nice to have: user should be able to pass a prio fee in here.
3. Probably should limit to x # of ATA instructions per tx and then iterate through multiple if needed...just to ensure we don't hit txn size/account size limits.

crates/lib/src/admin/token_util.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    let sig_bytes: [u8; 64] = signature
+        .bytes
+        .try_into()
+        .map_err(|_| KoraError::SigningError("Invalid signature length".to_string()))?;
+
+    let sig = Signature::from(sig_bytes);
+    tx.signatures = vec![sig];
+
+    match rpc_client.send_and_confirm_transaction(&tx).await {
```
amilz: I think since we're in a CLI, you could use `send_and_confirm_transaction_with_spinner` just cleaner ux.

crates/lib/src/admin/token_util.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut instructions = Vec::new();
+
+    // Check each token mint for existing ATA
+    for mint in &token_mints {
+        let ata = get_associated_token_address(&payment_address, mint);
+
+        match rpc_client.get_account(&ata).await {
+            Ok(_) => {
+                println!("✓ ATA already exists for token {mint}: {ata}");
+            }
```
amilz: WDYT about pulling some of this logic into a seperate function -- `find_missing_atas` -- you could then call it on `start` and do a warning log `Warning: ATAs not configured for xxx, yyy, zzz mints. Transactions may fail)`

crates/lib/src/token/interface.rs
```diff
@@ -9,6 +9,12 @@ use std::any::Any;
 
 use crate::validator::transaction_validator::TransactionValidator;
 
+pub struct DecodedTransfer {
+    pub amount: u64,
+    pub mint_index: Option<usize>,
+    pub destination_index: usize,
```
amilz: why not just return the `destination_key` DecodedTransfer? I think it's fine as-is, but I think you could save some steps in `crates/lib/src/token/token.rs`

crates/lib/src/token/token.rs
```diff
@@ -127,25 +130,53 @@ impl TokenUtil {
...
+    #[allow(clippy::too_many_arguments)]
     pub async fn process_token_transfer(
         ix: &CompiledInstruction,
         token_program: &dyn TokenInterface,
         account_keys: &[Pubkey],
         rpc_client: &RpcClient,
-        validation: &ValidationConfig,
         total_lamport_value: &mut u64,
         required_lamports: u64,
+        expected_destination: &Pubkey,
```
amilz: If I'm reading this right, this is the the Pubkey of the owner of the token account. I think SPL program uses `destination` often to refer to token accounts, so I'd suggest naming something like `expected_destination_owner` to avoid confusion.

tests/payment-address/main.rs
```diff
@@ -0,0 +1,14 @@
+// Payment address tests for Kora RPC server
+//
+// CONFIG: Uses tests/common/fixtures/paymaster-address-test.toml
```
amilz, line 3: ideal world, i think we'd have some type of global test config/env file that we set for tests e.g.--just would be a little easier to manage long term as we add more scenarios. but I think this is fine for now & well-documented. 
test type / kora.toml location / flag(s)

tests/payment-address/payment_address_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+async fn test_sign_transaction_if_paid_with_wrong_destination() {
+    let client = ClientTestHelper::get_test_client().await;
+    let rpc_client = RPCTestHelper::get_rpc_client().await;
+    let sender = SenderTestHelper::get_test_sender_keypair();
+    let wrong_destination = Keypair::new(); // Random wrong destination
+    let test_mint = USDCMintTestHelper::get_test_usdc_mint_pubkey();
+
+    // Create a transfer to the WRONG destination (not the payment address)
+    let sender_token_account = get_associated_token_address(&sender.pubkey(), &test_mint);
+    let wrong_dest_ata = get_associated_token_address(&wrong_destination.pubkey(), &test_mint);
```
amilz, line 80: I think you should probably do `get_or_create` here b/c the test might be failing just b/c wrong ata does not exist.

amilz (commented):
PR Looks good. Added a couple of suggestions. 

Pushed 2 commits
1. Docs update
2. TS Integration test fix

crates/cli/src/main.rs
```diff
@@ -47,6 +48,20 @@ enum ConfigCommands {
...
+#[derive(Subcommand)]
+enum RpcCommands {
+    /// Start the RPC server
+    Start {
+        #[command(flatten)]
+        rpc_args: Box<RpcArgs>,
+    },
+    /// Initialize ATAs for all allowed payment tokens
+    InitializeAtas {
+        #[command(flatten)]
```
amilz, line 68: imo `kora initialize-atas` is too much to type. I'd suggest `init` or `initialize` is fine. 
also `kora rpc initialize-atas -h` opens up the whole rpc help menu. would be nice to clean that up.

tests/payment-address/main.rs
dev-jodee: Yeah, I've dabbled with that, hot reloading configs, etc. but it becomes fairly complex, i.e. the AUTH configs, the middlewares aren't added to the server if we don't use it, so then we'd need to reload a new kora instance completely, maybe a restart or something, but it becomes involved a lot, we'lll have to dig deeper in test suites or something because right now its def not the best

crates/cli/src/main.rs
dev-jodee: Scared init could mean something else, there's other settings that we might init at some point, and people will type that once in a blue moon (unless they change their configs a lot of times), I think for clarity and since it's not used a lot it's fine to keep (people can always create aliases themselves)

makefiles/utils.mk
```diff
@@ -0,0 +1,101 @@
...
+# Start Kora server with flexible configuration
+# Usage: $(call start_kora_server,description,cargo_cmd,cargo_flags,config_file,setup_env)
+define start_kora_server
+	@$(call stop_kora_server)
+	@$(if $(5),\
+		printf "🔧 Setting up test environment...\n"; \
+		cargo run -p tests --bin setup_test_env;)
+	@echo "🚀 Starting Kora $(1)..."
+	@$(if $(2),\
+		$(2) -p kora-cli --bin kora $(3) -- --config $(4) --rpc-url $(TEST_RPC_URL) rpc start --private-key $(TEST_PRIVATE_KEY) --port $(TEST_PORT) &,\
```
amilz: Can we add quiet back to this? May need to make minor tweaks where it's invoked....but rn tests are filled with lots of server logs. 

```suggestion
		$(2) -p kora-cli --bin kora $(3) -- --config $(4) --rpc-url $(TEST_RPC_URL) rpc start --private-key $(TEST_PRIVATE_KEY) --port $(TEST_PORT) $(QUIET_OUTPUT) &,\
```

dev-jodee: Yeah I can add it back for now, but we'll have to have a better testing framework like I mentioned above, its getting a bit all over the place and quieting outputs also makes it hard to debug, etc. but for now it's fine

PR #175, opened 2025-08-18
feat: allow any spl paid token flag

Yolley:
Add `allow_any_spl_paid_token` flag that will bypass check (and takes precedence) for whether token is in `allowed_spl_paid_tokens` when checking whether transaction is paid.

For our use case we need to allow to use any SPL token for payment - the price of the token will be checked by the oracle. An alternative would be to allow `allowed_spl_paid_tokens` be empty, but I think that having this flag set explicitly makes more sense for backwards compatibility and security reasons.

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -58,13 +58,61 @@ impl Default for FeePayerBalanceMetricsConfig {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum SplTokenPaymentConfig {
+    All,
+    #[serde(untagged)]
+    Allowlist(Vec<String>),
+}
+
+impl Default for SplTokenPaymentConfig {
+    fn default() -> Self {
+        SplTokenPaymentConfig::Allowlist(vec![])
+    }
+}
+
+impl<'a> IntoIterator for &'a SplTokenPaymentConfig {
+    type Item = &'a String;
+    type IntoIter = std::slice::Iter<'a, String>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        match self {
+            SplTokenPaymentConfig::All => [].iter(),
+            SplTokenPaymentConfig::Allowlist(tokens) => tokens.iter(),
+        }
+    }
+}
+
+impl SplTokenPaymentConfig {
+    pub fn has_token(&self, token: &String) -> bool {
+        match self {
+            SplTokenPaymentConfig::All => true,
+            SplTokenPaymentConfig::Allowlist(tokens) => tokens.contains(token),
+        }
+    }
+
+    pub fn has_tokens(&self) -> bool {
+        match self {
+            SplTokenPaymentConfig::All => true,
+            SplTokenPaymentConfig::Allowlist(tokens) => !tokens.is_empty(),
+        }
+    }
+
+    pub fn as_slice(&self) -> &[String] {
+        match self {
+            SplTokenPaymentConfig::All => &[],
+            SplTokenPaymentConfig::Allowlist(v) => v.as_slice(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct ValidationConfig {
     pub max_allowed_lamports: u64,
     pub max_signatures: u64,
     pub allowed_programs: Vec<String>,
     pub allowed_tokens: Vec<String>,
-    pub allowed_spl_paid_tokens: Vec<String>,
+    pub allowed_spl_paid_tokens: SplTokenPaymentConfig,
     pub disallowed_accounts: Vec<String>,
     pub price_source: PriceSource,
     #[serde(default)] // Default for backward compatibility
@@ -81,7 +129,7 @@ impl ValidationConfig {
     }
 
     pub fn supports_token(&self, token: &str) -> bool {
-        self.allowed_spl_paid_tokens.iter().any(|s| s == token)
+        self.allowed_spl_paid_tokens.has_token(token)
     }
 }
 
@@ -387,7 +435,10 @@ mod tests {
         assert_eq!(config.validation.max_signatures, 10);
         assert_eq!(config.validation.allowed_programs, vec!["program1", "program2"]);
         assert_eq!(config.validation.allowed_tokens, vec!["token1", "token2"]);
-        assert_eq!(config.validation.allowed_spl_paid_tokens, vec!["token3"]);
+        assert_eq!(
+            config.validation.allowed_spl_paid_tokens,
+            SplTokenPaymentConfig::Allowlist(vec!["token3".to_string()])
+        );
         assert_eq!(config.validation.disallowed_accounts, vec!["account1"]);
         assert_eq!(config.validation.price_source, PriceSource::Jupiter);
         assert_eq!(config.kora.rate_limit, 100);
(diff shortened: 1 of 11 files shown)


Yolley:
@dev-jodee @amilz please take a look, I can't assign reviewers.

crates/lib/src/validator/config_validator.rs
```diff
@@ -669,6 +688,36 @@ mod tests {
...
+                fee_payer_policy: FeePayerPolicy::default(),
+                price: PriceConfig { model: PriceModel::Margin { margin: 0.1 } },
+            },
+            metrics: MetricsConfig::default(),
+            kora: KoraConfig::default(),
+        };
+
+        let _ = update_config(config);
+
+        let rpc_client = RpcClient::new("http://localhost:8899".to_string());
```
dev-jodee: Can we use mocked RPC client here for unit testing we don't want to require the local validator to be running.

crates/lib/src/token/token.rs
```diff
@@ -231,7 +231,9 @@ impl TokenUtil {
                     (token_state.mint(), actual_amount)
                 };
 
-                if !config.validation.allowed_spl_paid_tokens.contains(&mint_address.to_string()) {
+                if !config.validation.allow_any_spl_paid_token
```
dev-jodee: nit: have a function called is_spl_paid_token_valid(..mint..) { the 2 conditions } and use that func instead

crates/lib/src/validator/config_validator.rs
Yolley: Hm, I just used the same pattern with client initialisation as in other tests, will take a look.

dev-jodee: Hey, yeah might be the older version / so many changes have been happening, but here's the function to mock it https://github.com/solana-foundation/kora/pull/159/files#diff-e2aa8bf42a1b7b8111e7a9bc6b7990fa5efad819e5aa67fcdd2dbfed1ad79f53R93

Yolley: I see, thank you, updated the branch, please check.

crates/lib/src/token/token.rs
Yolley: Updated, added `is_spl_paid_token_valid` to the Validator as it's used both in the config itself during validation and in the TokenUtil.

amilz (commented):
Thanks for the PR @Yolley -- a couple of thoughts. 

## 1. Unified SplTokenPaymentConfig

Rather than two potentially conflicting fields in the config, could we do a single one that accepts an allowlist or ALL? I think it'd be a cleaner devex and reduce risk of conflicting configs.

Something like this:

```
  // In config.rs
  pub enum SplTokenPaymentConfig {
      All,                    // Accept any SPL token
      Allowlist(Vec<String>), // Accept only specific tokens
  }

  impl Default for SplTokenPaymentConfig {
      fn default() -> Self {
          SplTokenPaymentConfig::Allowlist(vec![])
      }
  }

  // In ValidationConfig
  pub struct ValidationConfig {
      // ... other fields ...
      pub allowed_spl_paid_tokens: SplTokenPaymentConfig,
  }
```

I think this change might require a tweak to `crates/lib/src/validator/config_validator.rs` to skip check if `All`. Could even add a warning if all is enabled: "Warning allowed for payments" 

## 2. Test coverage--would you mind adding a test or todo to make sure we throw if our oracle does not find a token.

Yolley:
> > @amilz I switched to the proposed solution - whilte I am not a fan of it, it's not necessarily bad. Also added a test for jup oracle for when it does not return price for a token. Please take look.
> 
> Hey @Yolley, sorry we've been crunching out features, the pr itself seems fine, I'm just trying to think of a case where a node operator would be willing to allow any spl tokens, with Solana having many 100s of thousands of tokens created every week or so, this seems extremely dangerous, what use case did you have in mind that would warrant allowing all spl tokens as payment?

Hey, our use-case is to use fee relayer when claiming an Airdrop - so sol/rent fees are covered by the relayer, and Airdrop may be created for any token on our platform.

Another use case I'd say could be for token-swaps? Like on a DEX - in case user does not have enough sol to cover processing fees, so the platform allows to pay with swapped tokens.

I don't see how it's "extremely" dangerous tho - we still check for the token price and make sure that the fees are covered according to pricing model configuration. I guess there could be an issue with prices changing a lot, but I don't see a clear fix for this right - I guess we could use a mix of volume/orderbook state or something, but it also depends on what Oracle can provide.

dev-jodee:
> @dev-jodee @amilz pretty please, I am tired of rebasing the branch.

Think the pr makes sense, sorry for the recurring rebase, branch is actively being worked on, finally merged the huge pre-release PR into main, if you rebase against main (might be easier to make a new branch from main and cherry-pick your commit into it), I can do final review and merge

crates/lib/src/tests/config_mock.rs
```diff
@@ -261,8 +262,8 @@ impl ValidationConfigBuilder {
         self
     }
 
-    pub fn with_allowed_spl_paid_tokens(mut self, tokens: Vec<String>) -> Self {
-        self.config.allowed_spl_paid_tokens = tokens;
+    pub fn with_allowed_spl_paid_tokens(mut self, config: SplTokenPaymentConfig) -> Self {
```
dev-jodee:
```suggestion
    pub fn with_allowed_spl_paid_tokens(mut self, spl_payment_config: SplTokenPaymentConfig) -> Self {
```

crates/lib/src/tests/toml_mock.rs
```diff
@@ -74,8 +77,8 @@ impl ConfigBuilder {
         self
     }
 
-    pub fn with_spl_paid_tokens(mut self, tokens: Vec<&str>) -> Self {
-        self.validation.allowed_spl_paid_tokens = tokens.iter().map(|s| s.to_string()).collect();
+    pub fn with_spl_paid_tokens(mut self, config: SplTokenPaymentConfig) -> Self {
```
dev-jodee:
```suggestion
    pub fn with_spl_paid_tokens(mut self, spl_payment_config: SplTokenPaymentConfig) -> Self {
```

crates/lib/src/config.rs
```diff
@@ -58,13 +58,61 @@ impl Default for FeePayerBalanceMetricsConfig {
...
+    fn into_iter(self) -> Self::IntoIter {
+        match self {
+            SplTokenPaymentConfig::All => [].iter(),
+            SplTokenPaymentConfig::Allowlist(tokens) => tokens.iter(),
+        }
+    }
+}
+
+impl SplTokenPaymentConfig {
+    pub fn has_token(&self, token: &str) -> bool {
```
dev-jodee, line 87: can we rename to has_allowed_payment_token for clarity

crates/lib/src/config.rs
```diff
@@ -58,13 +58,61 @@ impl Default for FeePayerBalanceMetricsConfig {
...
+        match self {
+            SplTokenPaymentConfig::All => true,
+            SplTokenPaymentConfig::Allowlist(tokens) => tokens.iter().any(|s| s == token),
+        }
+    }
+
+    pub fn has_tokens(&self) -> bool {
```
dev-jodee, line 94: nit: is_empty? rather than has_tokens

crates/lib/src/validator/config_validator.rs
dev-jodee: Use the RpcMockBuilder, it still points to a local url

crates/lib/src/config.rs
Yolley: `is_empty` to me seemed more ambiguous as technically it's empty when "all" is passed. So I'm specifically mention whether it has tokens, which also aligns with how `has_token` work.

crates/lib/src/config.rs
Yolley: Why tho? This method is specific to this config, config may not even be knowledgeable about whether it for allowed payment tokens or not. This config just stores token configuration, and either has specific tokens, or "all"

Yolley:
Hey, @amilz I've added a test `test_initialize_atas_when_all_tokens_are_allowed`, also regarding warning - even if user passes some tokens, node can still experience volatility if they pass volatile tokens, does it really make sense to issue another warning here, given that the configuration to enable all tokens is explicit?

crates/lib/src/config.rs
dev-jodee: Then it would make more sense to do SplTokenConfig and not SplTokenPaymentConfig if we want to make it has reusable as possible

Yolley: You're right, renamed it and pushed

PR #177, opened 2025-08-18
refactor: (PRO-213) Inner instruction support + refactor to have support of lookup tables and inner instructions across all of Kora

dev-jodee:
- Added `solana-transaction-status-client-types` dependency to `Cargo.toml` and `Cargo.lock`.
- Updated `fee.rs` to utilize `VersionedTransactionResolved` for transaction handling.
- Refactored transaction-related methods in `rpc_server` to work with the new transaction structure.
- Introduced `instruction_util.rs` for instruction parsing and utility functions.
- Removed deprecated `ix_utils.rs` as part of the cleanup.
- Adjusted tests to accommodate changes in transaction handling and ensure compatibility with new structures.

### DIFF
--- a/crates/lib/src/transaction/versioned_transaction.rs
+++ b/crates/lib/src/transaction/versioned_transaction.rs
@@ -18,7 +18,8 @@ use crate::{
     get_signer,
     state::get_config,
     transaction::{
-        instruction_util::IxUtils, ParsedSystemInstructionData, ParsedSystemInstructionType,
+        instruction_util::IxUtils, ParsedSPLInstructionData, ParsedSPLInstructionType,
+        ParsedSystemInstructionData, ParsedSystemInstructionType,
     },
     validator::transaction_validator::TransactionValidator,
     Signer,
@@ -38,6 +39,10 @@ pub struct VersionedTransactionResolved {
     // Parsed instructions by type (None if not parsed yet)
     parsed_system_instructions:
         Option<HashMap<ParsedSystemInstructionType, Vec<ParsedSystemInstructionData>>>,
+
+    // Parsed SPL instructions by type (None if not parsed yet)
+    parsed_spl_instructions:
+        Option<HashMap<ParsedSPLInstructionType, Vec<ParsedSPLInstructionData>>>,
 }
 
 impl Deref for VersionedTransactionResolved {
@@ -77,6 +82,7 @@ impl VersionedTransactionResolved {
             all_account_keys: vec![],
             all_instructions: vec![],
             parsed_system_instructions: None,
+            parsed_spl_instructions: None,
         };
 
         // 1. Resolve lookup table addresses based on transaction type
@@ -122,6 +128,7 @@ impl VersionedTransactionResolved {
                 transaction.message.static_account_keys(),
             ),
             parsed_system_instructions: None,
+            parsed_spl_instructions: None,
         }
     }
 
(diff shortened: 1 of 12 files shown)


crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -115,62 +101,132 @@ impl<'a> VersionedTransactionResolved<'a> {
...
 
-impl VersionedTransactionExt for VersionedTransaction {
-    fn get_all_account_keys(&self) -> Vec<Pubkey> {
-        self.message.static_account_keys().to_vec()
-    }
+        // 2. Fetch all instructions
+        let mut all_instructions =
+            IxUtils::uncompile_instructions(transaction.message.instructions(), &all_account_keys);
+
+        let inner_instructions = resolved.fetch_inner_instructions(rpc_client).await?;
```
amilz, line 113: nit. 
optional for instructions and keys...I think it might be easier to read if you do define `all_instructions` as an empty mutable vec then extend it with `inner` and `outer` rather than seeding it w/ the outer.

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -115,62 +101,132 @@ impl<'a> VersionedTransactionResolved<'a> {
...
+        resolved.all_instructions = all_instructions;
+
+        Ok(resolved)
     }
-}
 
-#[async_trait]
-pub trait VersionedTransactionUtilExt {
-    fn encode_b64_transaction(&self) -> Result<String, KoraError>;
+    /// Only use this is we built the transaction ourselves, because it won't do any checks for resolving LUT, etc.
```
amilz, line 121: should we prefix something like `INTERNAL_` ...? just to avoid makijng a mistake in the future?

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -115,62 +101,132 @@ impl<'a> VersionedTransactionResolved<'a> {
...
+            .await
+            .map_err(|e| KoraError::RpcError(format!("Failed to simulate transaction: {e}")))?;
 
-    async fn sign_and_send_transaction(
-        &self,
-        rpc_client: &RpcClient,
-    ) -> Result<(String, String), KoraError>;
+        if let Some(err) = simulation_result.value.err {
+            log::warn!(
+                "Transaction simulation failed: {err}, continuing without inner instructions",
```
amilz, line 147: should we just throw the failed simulation error instead?

crates/lib/src/rpc_server/method/transfer_transaction.rs
```diff
@@ -113,18 +115,22 @@ pub async fn transfer_transaction(
         Some(&fee_payer),
         &blockhash.0,
     ));
-    let mut transaction = TransactionUtil::new_unsigned_versioned_transaction(message);
+    let transaction = TransactionUtil::new_unsigned_versioned_transaction(message);
+
+    let mut simple_resolved_transaction =
```
amilz, line 120: imo just called `resolved_transaction`. Just dont' want to confuse folks that there's another tx type.

crates/lib/src/fee/fee.rs
```diff
@@ -38,24 +39,21 @@ impl FeeConfigUtil {
 
     async fn get_associated_token_account_creation_fees(
         rpc_client: &RpcClient,
-        transaction: &VersionedTransaction,
+        transaction: &VersionedTransactionResolved,
     ) -> Result<u64, KoraError> {
         const ATA_ACCOUNT_SIZE: usize = 165; // Standard ATA size
```
amilz, line 44: is this always true for token22? Aren't some token accounts larger?

dev-jodee: I believe it can be more, it'll be part of the token 2022 small rework !

crates/lib/src/transaction/versioned_transaction.rs
dev-jodee: renamed to something better

PR #179, opened 2025-08-19
feat: (PRO-212) token 2022 improvements

dev-jodee:


crates/lib/src/validator/config_validator.rs
amilz, line 1: I think we probably need a config validator that makes sure that the kora.toml inputs for extension names are valid and warn if invalid.

crates/lib/src/token/spl_token_2022.rs
```diff
@@ -191,149 +182,97 @@ impl TokenMint for Token2022Mint {
...
         }
-    }
 
-    /// Check if mint has metadata pointer
-    pub fn has_metadata_pointer(&self) -> bool {
-        self.has_mint_extension(ExtensionType::MetadataPointer)
-    }
+        // Apply transfer fees if configured
+        if let Some(fee) = self.calculate_transfer_fee(amount, current_epoch) {
+            return Ok(amount.saturating_sub(fee));
```
amilz, line 227: I'm probably just confusing myself, but do we really want to sub the transfer fee out of this? Wouldn't this put the Node on the hook for subsidizing the cost of the transfer fee?

amilz (approved):

crates/lib/src/token/spl_token_2022.rs
dev-jodee: Added a comment but tldr: we're removing the fee from the amount, because if user is "paying kora" 10 lamports, and the transfer fee is 1 lamports, then he's actually paying Kora 9 lamports

crates/lib/src/validator/config_validator.rs
dev-jodee: good idea!

PR #180, opened 2025-08-19
feat: (PRO-162) Implement Redis caching for account data

dev-jodee:
- Added Redis-based caching functionality to improve performance for Solana RPC calls.
- Introduced CacheUtil for managing cache operations, including initialization, retrieval, and invalidation of cached account data.
- Updated Kora configuration to support cache settings, including enabling/disabling and TTL values.
- Integrated caching into various components, replacing direct RPC calls with cache lookups where applicable.
- Enhanced Docker setup to include a Redis service for local development.
- Updated tests to validate caching behavior and configuration parsing.

### DIFF
--- a/crates/lib/src/rpc_server/method/transfer_transaction.rs
+++ b/crates/lib/src/rpc_server/method/transfer_transaction.rs
@@ -14,7 +14,7 @@ use crate::{
         TransactionUtil, VersionedMessageExt, VersionedTransactionOps, VersionedTransactionResolved,
     },
     validator::transaction_validator::TransactionValidator,
-    KoraError, Signer as _,
+    CacheUtil, KoraError, Signer as _,
 };
 
 #[derive(Debug, Deserialize, ToSchema)]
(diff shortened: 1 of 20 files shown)


crates/lib/src/rpc_server/method/transfer_transaction.rs
```diff
@@ -76,12 +76,11 @@ pub async fn transfer_transaction(
...
             token_program.get_associated_token_address(&destination, &token_mint.address());
 
-        let _ = rpc_client
-            .get_account(&source_ata)
+        CacheUtil::get_account_from_cache(rpc_client, &source_ata, false)
             .await
             .map_err(|_| KoraError::AccountNotFound(source_ata.to_string()))?;
 
-        if rpc_client.get_account(&dest_ata).await.is_err() {
+        if CacheUtil::get_account_from_cache(rpc_client, &dest_ata, false).await.is_err() {
```
amilz: Just a nit since we're calling this from `CacheUtil`, I think you could call this method `CacheUtil::get_account` -- it's implied that cache will be used if available. Could maybe even call it `CacheClient` since it's being used as a proxy for the rpc client. Up to you though.

amilz (approved):
Generally LGTM. Might be worth adding some config validation but NBD.

PR #182, opened 2025-08-19
chore:(PRO-224) fix ci

amilz:
- Update TS kora.toml path
- Update init ata for rust tests


amilz:
TS Integration Passing: https://github.com/solana-foundation/kora/actions/runs/17084668173

.github/workflows/rust.yml
```diff
@@ -2,14 +2,14 @@ name: Rust CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, release/pre-release-main-branch]
```
dev-jodee: can we do wildcard like release/* ?

PR #184, opened 2025-08-20
feat(signer): (PRO-215) implement multi-signer support with different strategies

dev-jodee:
- Added support for multiple signers in Kora gasless relayer, allowing for load balancing and high availability.
- Introduced `signers.toml` configuration file for defining signer pool and strategies.
- Updated CLI to accept `--signers-config` for multi-signer initialization.
- Refactored signer initialization logic to accommodate both single and multi-signer modes.
- Enhanced global state management to support signer pools.
- Added new types and methods for managing signer metadata and selection strategies.
- Updated relevant tests and documentation to reflect new functionality.

### DIFF
--- a/docs/contributors/ADDING_SIGNERS.md
+++ b/docs/contributors/ADDING_SIGNERS.md
@@ -16,7 +16,9 @@ Kora uses an enum-based architecture where all signers are wrapped in a unified 
     - Define types and configuration
     - Implement `KoraSigner`'s core signing methods (`sign` and `sign_solana`)
     - Add initialization logic based on your API's requirements
-- [ ] Update the `KoraSigner` enum and initialization logic
+- [ ] Update the `KoraSigner` enum
+- [ ] Update `SignerTypeConfig` enum in `config.rs` for multi-signer support
+- [ ] Add initialization logic in `init.rs`
 - [ ] Add CLI arguments
 - [ ] Update test scripts to include your signer (see below)
 - [ ] Update documentation to include your signer (see below)
@@ -242,7 +244,85 @@ impl super::Signer for KoraSigner {
 }
 ```
 
-### Step 5: Add Initialization Logic
+### Step 5: Update the SignerTypeConfig Enum
+
+Add your signer to the `SignerTypeConfig` enum in `crates/lib/src/signer/config.rs`:
+
+```rust
+/// Signer type-specific configuration with environment variable references
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum SignerTypeConfig {
+    // ... existing variants
+    
+    /// YourService signer configuration
+    YourService {
+        /// Environment variable for YourService API key
+        api_key_env: String,
+        /// Environment variable for YourService API secret
+        api_secret_env: String,
+        /// Environment variable for YourService wallet ID
+        wallet_id_env: String,
+    },
+}
+```
+
+Also update the `build_signer_from_config` method in the same file:
+
+```rust
+impl SignerConfig {
+    pub async fn build_signer_from_config(config: &SignerConfig) -> Result<KoraSigner, KoraError> {
+        match &config.config {
+            // ... existing cases
+            
+            SignerTypeConfig::YourService { api_key_env, api_secret_env, wallet_id_env } => {
+                let api_key = get_env_var_for_signer(api_key_env, &config.name)?;
+                let api_secret = get_env_var_for_signer(api_secret_env, &config.name)?;
+                let wallet_id = get_env_var_for_signer(wallet_id_env, &config.name)?;
+
+                let signer = YourServiceSigner::new(api_key, api_secret, wallet_id)
+                    .map_err(|e| {
+                        KoraError::ValidationError(format!(
+                            "Failed to create YourService signer '{}': {}",
+                            config.name, e
+                        ))
+                    })?;
+
+                Ok(KoraSigner::YourService(signer))
+            }
+        }
+    }
+}
+```
+
+And update the validation method:
+
+```rust
+fn validate_individual_signer_config(&self, index: usize) -> Result<(), KoraError> {
+    match &self.config {
+        // ... existing cases
+        
+        SignerTypeConfig::YourService { api_key_env, api_secret_env, wallet_id_env } => {
+            let env_vars = [
+                ("api_key_env", api_key_env),
+                ("api_secret_env", api_secret_env),
+                ("wallet_id_env", wallet_id_env),
+            ];
+            for (field_name, env_var) in env_vars {
+                if env_var.is_empty() {
+                    return Err(KoraError::ValidationError(format!(
+                        "YourService signer '{}' must specify non-empty {}",
+                        self.name, field_name
+                    )));
+                }
+            }
+        }
+    }
+    Ok(())
+}
+```
+
+### Step 6: Add Initialization Logic
 
 Update `crates/lib/src/signer/init.rs` to include your signer:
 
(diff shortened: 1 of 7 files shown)


crates/lib/src/metrics/balance.rs
```diff
@@ -1,65 +1,101 @@
...
+            // Batch fetch all signer balances
+            for signer_info in &signers_info {
+                let pubkey = Pubkey::from_str(&signer_info.public_key).map_err(|e| {
+                    KoraError::InternalServerError(format!(
+                        "Invalid signer pubkey {}: {e}",
+                        signer_info.public_key
+                    ))
+                })?;
+
+                match CacheUtil::get_account(rpc_client, &pubkey, false).await {
```
amilz, line 67: This sequence is probably fine, but we might need to prepare for allowing more flexibility for this kind of stuff to avoid RPC rate limits.

crates/lib/src/rpc_server/method/estimate_transaction_fee.rs
amilz, line 1: I'm not convinced it makes sense to include this arg/response from this method...since it's nothing to do w/ signing. we can still validate that they have a signer in the txn for accounting, but I dont' think it should matter here.

crates/lib/src/signer/config.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    /// Validate an individual signer configuration
+    fn validate_individual_signer_config(&self, index: usize) -> Result<(), KoraError> {
+        if self.name.is_empty() {
+            return Err(KoraError::ValidationError(format!(
+                "Signer at index {index} must have a non-empty name"
+            )));
+        }
+
+        // Validate fields based on signer type
```
amilz: Should this be a signer impl so that this logic lives in `crates/lib/src/signer/{signer-type}`? and then down below we call some `.validate_individual_config()` or something like that?

crates/lib/src/state.rs
```diff
@@ -5,34 +5,107 @@ use std::sync::{
...
 // Global config with zero-cost reads and hot-reload capability
 static GLOBAL_CONFIG: AtomicPtr<Config> = AtomicPtr::new(std::ptr::null_mut());
 
-/// Initialize the global signer with a KoraSigner instance
-pub fn init_signer(signer: KoraSigner) -> Result<(), KoraError> {
-    let mut signer_guard = GLOBAL_SIGNER.write();
-    if signer_guard.is_some() {
-        return Err(KoraError::InternalServerError("Signer already initialized".to_string()));
+/// Get a request-scoped signer that should be used consistently throughout a single RPC request
+/// This advances the round-robin counter once per request for consistent signer usage
```
amilz: I don't see the advance logic in here?

crates/lib/src/state.rs
```diff
@@ -5,34 +5,107 @@ use std::sync::{
...
+pub fn get_signer_pool() -> Result<Arc<SignerPool>, KoraError> {
+    let pool_guard = GLOBAL_SIGNER_POOL.read();
+    match &*pool_guard {
+        Some(pool) => Ok(Arc::clone(pool)),
+        None => Err(KoraError::InternalServerError("Signer pool not initialized".to_string())),
     }
 }
 
+/// Get the next signer from the pool with error handling and metrics tracking
+pub fn get_next_signer_with_tracking() -> Result<(Arc<KoraSigner>, String), KoraError> {
```
amilz: I think this is unused?

crates/lib/src/signer/pool.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /// Weighted selection strategy (weighted random)
+    fn weighted_select(&self) -> Result<&SignerWithMetadata, KoraError> {
+        use rand::Rng;
+
+        let total_weight: u32 = self.signers.iter().map(|s| s.weight).sum();
+        if total_weight == 0 {
+            return Err(KoraError::InternalServerError("All signers have zero weight".to_string()));
+        }
```
amilz: This should be static, right? IMO just store it in the `SignerPool`

crates/lib/src/config.rs
amilz, line 1: FWIW, I'm inclined to make `payment_address` required now. Accidentally leaving this off when you have a bunch of Signers could be really annoying/costly.

crates/cli/src/main.rs
amilz, line 1: RN I think the`initialize_paymaster_atas` via `find_missing_atas` is only finding the "next" the signer wallets (meaning when we initialize, we will not initialize ATAs for subsequent signers. I think it just calls `get_request_signer` 1x.

We need to make sure our test coverage is testing subsequent signers.

tests/multi-signers/multi_signer_tests.rs
amilz, line 1: I think we need to add tests sending series of transactions to test the sequence and ATA initalization

amilz (approved):
Good PR.
See my comments. 2 issues I'm most concerned about: 

1. Determining `payment_address` if one is not specified (IMO we should probably require the field now). Otherwise need to handle initializing ATAs for all wallets in the Pool.
2. Config validator should probably validate the signers file (and maybe ideally do a ping to the respective signer networks)
3. Devex. I think the signer _hint_ thing could be a little confusing for users. Making sure i understand: a user can specify any signer from the config using a "hint", otherwise Kora will choose one using the signer strategy and return it to the user. IMO, let's just call it what it is: `kora_signer` (instead of hint). The user is just telling the server which signer it wants to use. 

I added a PR to add docs & some minor updates to SDK: #185

crates/lib/src/rpc_server/method/estimate_transaction_fee.rs
dev-jodee: the only reason is that we have checks for fee payer outflow, etc and if we want to go over all the signers it might be pretty slow if the kora node has a lot of signers

dev-jodee: and all of them are optionnal

tests/multi-signers/multi_signer_tests.rs
dev-jodee: will do in our overall test pr :)

crates/lib/src/state.rs
dev-jodee: yeah metrics werent used for now, was debating keeping it or removing and just adding back when we put support, i'll clean it

crates/lib/src/metrics/balance.rs
dev-jodee: We'll most likely need to support multi rpc, etc, this is v0.1 of the cache to at least mitigate some of the calls to RPCs

crates/lib/src/state.rs
dev-jodee: will remove, keep code clean and only for what we use

crates/lib/src/signer/pool.rs
dev-jodee: added as a param to the pool to only calculate once and check if weighted with weight = 0

crates/lib/src/config.rs
dev-jodee: We can talk about it, but I tend to prefer to be on the side of letting people config what they want, instead of enforcing something on them. Hard to think of all the use cases, there's a potential where some node operators could want the used signer to be paid directly i.e.

crates/cli/src/main.rs
dev-jodee: Yeah I'll make it possible to loop or specify which signer

dev-jodee:
> Good PR. See my comments. 2 issues I'm most concerned about:
> 
> 1. Determining `payment_address` if one is not specified (IMO we should probably require the field now). Otherwise need to handle initializing ATAs for all wallets in the Pool.
> 2. Config validator should probably validate the signers file (and maybe ideally do a ping to the respective signer networks)
> 3. Devex. I think the signer _hint_ thing could be a little confusing for users. Making sure i understand: a user can specify any signer from the config using a "hint", otherwise Kora will choose one using the signer strategy and return it to the user. IMO, let's just call it what it is: `kora_signer` (instead of hint). The user is just telling the server which signer it wants to use.
> 
> I added a PR to add docs & some minor updates to SDK: #185

Will rename, hint is common for certain types of services but I think you're right it might be confusing here. Answered for payment address on the comment itself. Config validation it'll crash if it doesn't follow what we expect, but yeah we could add checking on them, ill add as a task.

crates/lib/src/signer/config.rs
dev-jodee: Hmmmm, yeah I guess we could split it in different signers

PR #190, opened 2025-08-21
feat: unit testing coverage

dev-jodee:
> [!IMPORTANT]
> Enhance unit testing coverage for token, transaction, and validator modules, adding tests for various functions and scenarios.
> 
>   - **Token Module**:
>     - Add tests for `TokenType` methods in `token.rs` to verify correct token program retrieval.
>     - Add tests for `TokenUtil` methods to validate token price calculations and extension validations.
>   - **Transaction Module**:
>     - Add tests for `VersionedMessageExt` in `versioned_message.rs` to ensure base64 encoding correctness.
>     - Add tests for `VersionedTransactionResolved` in `versioned_transaction.rs` to verify transaction resolution and signing.
>   - **Validator Module**:
>     - Add tests for `AccountType` validation in `account_validator.rs` to ensure correct account type handling.
>     - Add tests for `validate_account` function to check account validation logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fkora&utm_source=github&utm_medium=referral)<sup> for 5a528f17bf8066f735b589b0a43ed21ab3d0a3d0. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## 📊 Test Coverage
Coverage

**Coverage: 85.8%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/17237454868)

### DIFF
--- a/crates/lib/src/metrics/balance.rs
+++ b/crates/lib/src/metrics/balance.rs
@@ -277,8 +277,7 @@ mod tests {
             .build_and_setup();
 
         let result = BalanceTracker::init().await;
-        // May succeed or fail depending on if already initialized, both are valid
-        assert!(result.is_ok() || result.is_err());
+        assert!(result.is_ok());
     }
 
     #[tokio::test]
--- a/crates/lib/src/metrics/balance.rs
+++ b/crates/lib/src/metrics/balance.rs
@@ -1,8 +1,8 @@
-use crate::{
-    cache::CacheUtil,
-    error::KoraError,
-    state::{get_config, get_signers_info},
-};
+#[cfg(not(test))]
+use crate::state::get_config;
+#[cfg(test)]
+use crate::tests::config_mock::mock_state::get_config;
+use crate::{cache::CacheUtil, error::KoraError, state::get_signers_info};
 use prometheus::{register_gauge_vec, GaugeVec};
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_sdk::pubkey::Pubkey;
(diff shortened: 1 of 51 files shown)


crates/lib/src/signer/memory_signer/solana_signer.rs
```diff
@@ -78,3 +78,132 @@ impl SolanaMemorySigner {
...
+    }
+
+    #[test]
+    fn test_from_private_key_string_base58() {
+        let _m = ConfigMockBuilder::new().build_and_setup();
+
+        let test_private_key = "5MaiiCavjCmn9Hs1o3eznqDEhRwxo7pXiAYez7keQUviUkauRiTMD8DrESdrNjN8zd9mTmVhRvBJeg5vhyvgrAhG";
+        let signer = SolanaMemorySigner::from_private_key_string(test_private_key).unwrap();
+
+        // Just verify we can create the signer - actual pubkey depends on the private key
```
amilz: since you are using a specific private key, might as well check that it's initializing correct (`5pVyoAeURQHNMVU7DmfMHvCDNmTEYXWfEwc136GYhTKG`)?

crates/lib/src/signer/privy/signer.rs
```diff
@@ -118,3 +118,251 @@ impl PrivySigner {
...
+
+    #[test]
+    fn test_get_auth_header() {
+        let signer = PrivySigner::new(
+            "test_app".to_string(),
+            "test_secret".to_string(),
+            "wallet123".to_string(),
+        );
+
+        let auth_header = signer.get_auth_header();
```
amilz: just realized going through this that we only have `get_auth_header` on privy--it was confusing for me looking at this b/c i was thinking this was referring to our auth config (api key/hmac). wondering if we rename this `get_privy_auth_header` to avoid confusion. and probably should make sure we explicitly add tests for hmac + privy to ensure there's no conflicting/overriding header issues.

crates/lib/src/token/token.rs
```diff
@@ -344,58 +349,124 @@ impl TokenUtil {
...
+        let mixed_tokens = vec![WSOL_DEVNET_MINT.to_string(), "invalid_address".to_string()];
+        let result = TokenUtil::check_valid_tokens(&mixed_tokens);
+        assert!(result.is_err());
+        assert!(matches!(result.unwrap_err(), KoraError::ValidationError(_)));
+    }
+
+    #[tokio::test]
+    async fn test_get_mint_valid() {
+        let _lock = ConfigMockBuilder::new().build_and_setup();
+        let mint = Pubkey::from_str(WSOL_DEVNET_MINT).unwrap();
```
amilz, line 436: IMO these mint tests are a bit misleading/tricky to follow. Maybe it's fine...but if I'm tracing the calls back to the mock...the mock rpc is indifferent to the mint address. I think just a comment like `// any mint will work for testing the mock data` (or something like that) would be helpful in reading this in the future.

crates/lib/src/rpc_server/method/get_config.rs
```diff
@@ -36,3 +36,27 @@ pub async fn get_config() -> Result<GetConfigResponse, KoraError> {
...
+        let result = get_config().await;
+
+        assert!(result.is_ok(), "Should successfully get config");
+        let response = result.unwrap();
+        assert!(!response.fee_payers.is_empty(), "Should have at least one fee payer");
+        assert!(!response.fee_payers[0].is_empty(), "Fee payer pubkey should not be empty");
+        assert!(
+            response.validation_config.max_allowed_lamports > 0,
+            "Should have max_allowed_lamports set"
+        );
```
amilz: nit: could probably be more specific w/ the assertions here to match the default mock config

crates/lib/src/rpc_server/method/sign_and_send_transaction.rs
```diff
@@ -44,3 +44,49 @@ pub async fn sign_and_send_transaction(
...
+            signer_key: Some("invalid_pubkey".to_string()),
+        };
+
+        let result = sign_and_send_transaction(&rpc_client, request).await;
+
+        assert!(result.is_err(), "Should fail with invalid signer key");
+        let error = result.unwrap_err();
+        assert!(matches!(error, KoraError::ValidationError(_)), "Should return ValidationError");
+    }
+}
```
amilz, line 92: Could use mockrpc to mock a successful sign & send

crates/lib/src/rpc_server/method/estimate_transaction_fee.rs
amilz, line 1: Happy path unit tests are missing for a few methods: `sign_transaction`, `sign_and_send_transaction`, `estimate_transaction_fee`, and `transfer_transaction`.

crates/lib/src/rpc_server/method/get_config.rs
amilz, line 1: since this this method actually gets the config--it could make sense to mock a few variants of config (e.g., `get_default_config_with_payment_address`)

crates/lib/src/metrics/balance.rs
```diff
@@ -149,108 +149,248 @@ impl BalanceTracker {
...
+                    .with_fee_payer_balance(FeePayerBalanceMetricsConfig {
+                        enabled: true,
+                        expiry_seconds: 30,
+                    })
+                    .build(),
+            )
+            .build_and_setup();
+
+        let result = BalanceTracker::init().await;
+        // May succeed or fail depending on if already initialized, both are valid
```
amilz: shouldn't we have 2 test cases then? 1 when initialized and why when not already initialized? otherwise idk if there's anyway this test could ever fail

crates/lib/src/signer/privy/signer.rs
dev-jodee: Renamed, but for the overriding issues, those are completely separate http calls, with their own set of headers, and privy's headers are generated by us, don't see how it could be an issue in relation to our own headers.

crates/lib/src/rpc_server/method/sign_and_send_transaction.rs
dev-jodee: same answer as the other one :)

crates/lib/src/rpc_server/method/estimate_transaction_fee.rs
dev-jodee: It's actually pretty complex to test the happy path because of the way the transaction is built from "from_transaction" and there are internal rpc calls, for now I think we're fine with none since we have the integration tests, but definitely something that could be good to implement in the future.

PR #212, opened 2025-08-29
feat: (PRO-268)  Enhance fee estimation with transfer fee calculation…

dev-jodee:
…s and Kora's price model. Also removed check for non transferable and cpi guard, as they will fail at the transaction level instead.

- Introduced `calculate_transfer_fees` method to compute transfer fees for token transactions.
- Updated `estimate_transaction_fee` to return a `TotalFeeCalculation` struct, encapsulating all fee components.
- Removed forced non transferable and cpi guard check, will be validated by the chain when the transaction is submitted.

----

> [!IMPORTANT]
> Enhances fee estimation by adding transfer fee calculation and updating the fee structure, while removing certain checks now handled at the transaction level.
> 
>   - **Behavior**:
>     - Adds `calculate_transfer_fees` in `fee.rs` to compute transfer fees for token transactions.
>     - Updates `estimate_transaction_fee` to return `TotalFeeCalculation` struct with detailed fee components.
>     - Removes non-transferable and CPI guard checks, now validated by the chain.
>   - **Fee Calculation**:
>     - Introduces `TotalFeeCalculation` struct in `fee.rs` to encapsulate fee components.
>     - Updates `estimate_kora_fee` to apply Kora's price model.
>     - Adds `calculate_fee_in_token` to convert fees to specific tokens.
>   - **Tests**:
>     - Updates tests in `fee.rs` and `estimate_transaction_fee.rs` to reflect new fee calculation logic.
>     - Adds tests for transfer fee calculation and fee estimation with uninitialized ATAs.
>   - **Documentation**:
>     - Adds `FEES.md` to document fee estimation components and calculation methods.
>     - Updates `README.md` to include fee estimation guide reference.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fkora&utm_source=github&utm_medium=referral)<sup> for 25f9d0a1266812301f2192124b8d31d8cd7e5bba. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## 📊 Test Coverage
Coverage

**Coverage: 86.3%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/17414539047)

### DIFF
--- a/crates/lib/src/fee/fee.rs
+++ b/crates/lib/src/fee/fee.rs
@@ -165,12 +165,11 @@ impl FeeConfigUtil {
     async fn calculate_transfer_fees(
         rpc_client: &RpcClient,
         transaction: &mut VersionedTransactionResolved,
+        fee_payer: &Pubkey,
     ) -> Result<u64, KoraError> {
         let config = get_config()?;
 
-        // Get fee payer and payment destination before parsing instructions to avoid borrow conflicts
-        let fee_payer = transaction.transaction.message.static_account_keys()[0];
-        let payment_destination = config.kora.get_payment_address(&fee_payer)?;
+        let payment_destination = config.kora.get_payment_address(fee_payer)?;
 
         let parsed_spl_instructions = transaction.get_or_parse_spl_instructions()?;
 
@@ -192,7 +191,6 @@ impl FeeConfigUtil {
                     match CacheUtil::get_account(rpc_client, destination_address, false).await {
                         Ok(account) => account,
                         Err(_) => {
-                            // Destination doesn't exist, can't be a payment to Kora, skip this instruction
                             continue;
                         }
                     };
@@ -204,7 +202,7 @@ impl FeeConfigUtil {
                     destination_token_program.unpack_token_account(&destination_account.data)?;
 
                 if destination_token_account.owner() != payment_destination {
-                    // This is not a payment transaction, skip it
+                    // This is not a transfer transaction to Kora, skip it
                     continue;
                 }
 
@@ -275,7 +273,7 @@ impl FeeConfigUtil {
         };
 
         let transfer_fee_config_amount =
-            FeeConfigUtil::calculate_transfer_fees(rpc_client, transaction).await?;
+            FeeConfigUtil::calculate_transfer_fees(rpc_client, transaction, fee_payer).await?;
 
         let total_fee_lamports = base_fee
             + account_creation_fee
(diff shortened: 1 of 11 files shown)


crates/lib/src/fee/fee.rs
```diff
@@ -149,12 +161,87 @@ impl FeeConfigUtil {
...
+                // Skip if destination account doesn't exist (not a payment to existing Kora account)
+                let destination_account =
+                    match CacheUtil::get_account(rpc_client, destination_address, false).await {
+                        Ok(account) => account,
+                        Err(_) => {
+                            continue;
+                        }
+                    };
+
+                let destination_token_program =
```
amilz: we perform a lot of this same logic in `has_payment_instruction` -- you might be able to create a helper function to streamline?

crates/lib/src/fee/fee.rs
```diff
@@ -185,11 +272,58 @@ impl FeeConfigUtil {
...
+        is_payment_required: bool,
+        price_source: Option<PriceSource>,
+    ) -> Result<TotalFeeCalculation, KoraError> {
+        // Get the raw transaction fees
+        let mut fee_calculation =
+            Self::estimate_transaction_fee(rpc_client, transaction, fee_payer, is_payment_required)
+                .await?;
+
+        // Apply Kora's price model
+        if let Some(price_source) = price_source {
```
amilz, line 354: WDYT about refactoring a little so we can avoid a compute/rpc heavy `estimate_transaction_fee` call for free/fixed payment types? Unforutnately I think this means we'd need something different than `get_required_lamports` since it requires `fee_calculation` beforehand

crates/lib/src/rpc_server/method/estimate_transaction_fee.rs
amilz, line 1: wdyt about moving the `fee_in_token` logic here to `FeeConfigUtil`?

crates/lib/src/token/token.rs
```diff
@@ -146,29 +146,27 @@ impl TokenUtil {
...
         source_address: &Pubkey,
         destination_address: &Pubkey,
-        mint: &Option<Pubkey>,
+        mint: &Pubkey,
     ) -> Result<(), KoraError> {
         let config = &get_config()?.validation.token_2022;
 
         let token_program = Token2022Program::new();
 
         // Get mint account data and validate mint extensions (force refresh in case extensions are added)
```
amilz, line 155: curious if these checks would hold up if a token or mint account is initialized w/ an invalid extension in this step.

PR #228, opened 2025-09-29
feat: (PRO-345) Fixed inner instructions

dev-jodee:
- Inner instructions weren't properly being handled
- Added parsing of Parsed / PartiallyDecoded instructions coming back from the RPC
- Added tests with inner instructions to make sure the inner instructions are properly used

----

> [!IMPORTANT]
> Enhanced inner instruction handling by adding parsing for `Parsed` and `PartiallyDecoded` instructions, with comprehensive tests and updated dependencies.
> 
>   - **Behavior**:
>     - Added parsing for `Parsed` and `PartiallyDecoded` instructions in `instruction_util.rs`.
>     - Updated `fetch_inner_instructions()` in `versioned_transaction.rs` to handle new instruction types.
>     - Added tests for inner instruction parsing in `instruction_util.rs` and `versioned_transaction.rs`.
>   - **Dependencies**:
>     - Added `solana-transaction-status` to `Cargo.toml` and `Cargo.lock`.
>     - Updated `spl-associated-token-account` version in `Cargo.lock`.
>   - **Tests**:
>     - Added comprehensive tests in `fee_estimation.rs` for transaction fee estimation with various scenarios.
>     - Updated test runner and transaction builder in `test_runner.rs` and `transaction.rs` to support new test cases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fkora&utm_source=github&utm_medium=referral)<sup> for ce2478a4df99bd688cbe57e12e5071138ac48e71. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 85.4%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/18135311608)

### DIFF
--- a/crates/lib/src/transaction/instruction_util.rs
+++ b/crates/lib/src/transaction/instruction_util.rs
@@ -210,6 +210,10 @@ impl IxUtils {
         Some(ix.accounts[index].pubkey)
     }
 
+    pub fn build_default_compiled_instruction(program_id_index: u8) -> CompiledInstruction {
+        CompiledInstruction { program_id_index, accounts: vec![], data: vec![] }
+    }
+
     pub fn uncompile_instructions(
         instructions: &[CompiledInstruction],
         account_keys: &[Pubkey],
@@ -260,30 +264,21 @@ impl IxUtils {
             }
             UiInstruction::Parsed(ui_parsed) => match ui_parsed {
                 UiParsedInstruction::Parsed(parsed) => {
+                    let account_keys_hashmap = Self::build_account_keys_hashmap(all_account_keys);
                     // Reconstruct based on program type
                     if parsed.program_id == SYSTEM_PROGRAM_ID.to_string() {
-                        let account_keys_hashmap =
-                            Self::build_account_keys_hashmap(all_account_keys);
                         Self::reconstruct_system_instruction(parsed, &account_keys_hashmap).ok()
                     } else if parsed.program == spl_token::ID.to_string()
                         || parsed.program == spl_token_2022::ID.to_string()
                     {
-                        let account_keys_hashmap =
-                            Self::build_account_keys_hashmap(all_account_keys);
                         Self::reconstruct_spl_token_instruction(parsed, &account_keys_hashmap).ok()
                     } else {
                         // For unsupported programs, create a stub instruction with just the program ID
                         // This ensures the program ID is preserved for security validation
                         let program_id = parsed.program_id.parse::<Pubkey>().ok()?;
-                        let account_keys_hashmap =
-                            Self::build_account_keys_hashmap(all_account_keys);
                         let program_id_index = *account_keys_hashmap.get(&program_id)?;
 
-                        Some(CompiledInstruction {
-                            program_id_index,
-                            accounts: vec![],
-                            data: vec![],
-                        })
+                        Some(Self::build_default_compiled_instruction(program_id_index))
                     }
                 }
                 UiParsedInstruction::PartiallyDecoded(partial) => {
@@ -506,10 +501,7 @@ impl IxUtils {
             }
             _ => {
                 log::error!("Unsupported system instruction type: {}", instruction_type);
-                Err(KoraError::SerializationError(format!(
-                    "Unsupported system instruction type: {}",
-                    instruction_type
-                )))
+                Ok(Self::build_default_compiled_instruction(program_id_index))
             }
         }
     }
@@ -722,10 +714,7 @@ impl IxUtils {
             }
             _ => {
                 log::error!("Unsupported token instruction type: {}", instruction_type);
-                Err(KoraError::SerializationError(format!(
-                    "Unsupported SPL token instruction type: {}",
-                    instruction_type
-                )))
+                Ok(Self::build_default_compiled_instruction(program_id_index))
             }
         }
     }


Cargo.toml
```diff
@@ -40,6 +40,7 @@ solana-commitment-config = "2.1.9"
 solana-message = "2.1.9"
 solana-system-interface = "1.0.0"
 solana-transaction-status-client-types = "2.1.9"
+solana-transaction-status = "2.1.9"
```
amilz, line 43: Maybe not compatible but i think there's `2.3.11`

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -120,6 +163,466 @@ impl IxUtils {
             .collect()
     }
 
+    /// Reconstruct a CompiledInstruction from various UiInstruction formats
```
amilz, line 240: wdyt about adding a little extra context here so future folks know why it's necessary
and maybe link to [agave source](https://github.com/anza-xyz/agave/blob/68032b576dc4c14b31c15974c6734ae1513980a3/transaction-status/src/parse_system.rs#L94)

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -120,6 +163,466 @@ impl IxUtils {
...
+                    // Reconstruct based on program type
+                    if parsed.program_id == SYSTEM_PROGRAM_ID.to_string() {
+                        Self::reconstruct_system_instruction(parsed, all_account_keys)
+                    } else if parsed.program == spl_token::ID.to_string()
+                        || parsed.program == spl_token_2022::ID.to_string()
+                    {
+                        Self::reconstruct_spl_token_instruction(parsed, all_account_keys)
+                    } else {
+                        log::error!("Unsupported parsed program: {}", parsed.program);
+                        None
```
amilz: Are there any risks with returning `None` here when we know there's some unknown ix data? 
Just looking at [Agave](https://github.com/anza-xyz/agave/tree/68032b576dc4c14b31c15974c6734ae1513980a3/transaction-status/src), there are a few other programs with parsed ix. Maybe should add ata

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -120,6 +163,466 @@ impl IxUtils {
...
+                            return Some(CompiledInstruction {
+                                program_id_index: program_idx as u8,
+                                accounts: account_indices,
+                                data: bs58::decode(&partial.data).into_vec().unwrap_or_default(),
+                            });
+                        }
+                    }
+
+                    log::error!("Failed to reconstruct partially decoded instruction");
+                    None
```
amilz, line 310: Same Q as above

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -120,6 +163,466 @@ impl IxUtils {
...
+
+                Some(CompiledInstruction {
+                    program_id_index,
+                    accounts: vec![source_idx, destination_idx],
+                    data,
+                })
+            }
+            PARSED_DATA_FIELD_CREATE_ACCOUNT => {
+                let source =
+                    info.get(PARSED_DATA_FIELD_SOURCE)?.as_str()?.parse::<Pubkey>().ok()?;
```
amilz: Since there's a lot `?.as_str()?`..., WDYT about something like this to `IxUtils` for tigthening up the code a little +  error handling? maybe overkill...

```rust
    fn get_field_as_str<'a>(
        info: &'a serde_json::Value,
        field_name: &str,
    ) -> Result<&'a str, KoraError> {
        info.get(field_name)
            .ok_or_else(|| {
                KoraError::MalformedParsedData(format!("Missing field '{}'", field_name))
            })?
            .as_str()
            .ok_or_else(|| {
                KoraError::MalformedParsedData(format!("Field '{}' is not a string", field_name))
            })
    }

    fn get_field_as_pubkey(
        info: &serde_json::Value,
        field_name: &str,
    ) -> Result<Pubkey, KoraError> {
        let pubkey_str = Self::get_field_as_str(info, field_name)?;
        pubkey_str.parse::<Pubkey>().map_err(|e| {
            KoraError::MalformedParsedData(format!(
                "Field '{}' is not a valid pubkey: {}",
                field_name, e
            ))
        })
    }

// other types as needed
```

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -120,6 +163,466 @@ impl IxUtils {
...
+            },
+        }
+    }
+
+    /// Reconstruct system program instructions from parsed format
+    fn reconstruct_system_instruction(
+        parsed: &solana_transaction_status_client_types::ParsedInstruction,
+        all_account_keys: &[Pubkey],
+    ) -> Option<CompiledInstruction> {
+        let program_id_index = all_account_keys.iter().position(|k| k == &SYSTEM_PROGRAM_ID)? as u8;
```
amilz: looks like we do a lot of `all_account_keys.iter()`
would it make sense to iter once and make a `HashMap<Pubkey, usize>` and then call `let target_index = account_map.get(&target)?;` or something like that?

Cargo.toml
dev-jodee: Yeah trying to keep the same version across the different packages, might make another pr where I try to update all of them, don't think it'd be too much

crates/lib/src/transaction/instruction_util.rs
dev-jodee: Talked with you on Slack

-> Local tests were just an issue on your side
-> Supporting ATA, we don't have any filter on ATA actions and creation is picked up by System Program so we don't need to parse it
-> Erroring out will be an issue because then we need to support ALL parsed instructions (including future ones that would break Kora)

crates/lib/src/transaction/instruction_util.rs
dev-jodee: answered above !

crates/lib/src/transaction/instruction_util.rs
dev-jodee: yesss

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -181,21 +262,34 @@ impl IxUtils {
                 UiParsedInstruction::Parsed(parsed) => {
                     // Reconstruct based on program type
                     if parsed.program_id == SYSTEM_PROGRAM_ID.to_string() {
-                        Self::reconstruct_system_instruction(parsed, all_account_keys)
+                        let account_keys_hashmap =
```
amilz: can we define the hashmap before the if or before the match even and just have it 1x?

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -229,205 +321,226 @@ impl IxUtils {
...
+                Ok(CompiledInstruction {
                     program_id_index,
                     accounts: vec![nonce_account_idx, recipient_idx, nonce_authority_idx],
                     data,
                 })
             }
             _ => {
                 log::error!("Unsupported system instruction type: {}", instruction_type);
-                None
+                Err(KoraError::SerializationError(format!(
```
amilz: did you want to follow this pattern for System & Token Program too?

```
                        Some(CompiledInstruction {
                            program_id_index,
                            accounts: vec![],
                            data: vec![],
                        })
```

crates/lib/src/transaction/instruction_util.rs
dev-jodee: didn't want to build the hashmap for the "else" case for no reason, since we don't care about those programs

dev-jodee: nvm I get what you mean lol I use it in the else too

amilz (approved):

PR #233, opened 2025-10-14
feat: (PRO-413) Enhance fee payer outflow calculation to include SPL …

dev-jodee:
…token transfers

- Updated `calculate_fee_payer_outflow` to account for SPL token transfers, improving accuracy in fee estimation.
- Added comprehensive tests for SPL token transfer scenarios to ensure correct fee calculations.
- Adjusted transaction validation to utilize updated outflow calculations, ensuring consistency across transaction processing.

----

> [!IMPORTANT]
> Enhance fee payer outflow calculation to include SPL token transfers for accurate fee estimation and update related transaction validation and tests.
> 
>   - **Behavior**:
>     - Update `calculate_fee_payer_outflow` in `fee.rs` to include SPL token transfers for accurate fee estimation.
>     - Adjust transaction validation in `transaction_validator.rs` to use updated outflow calculations.
>   - **Testing**:
>     - Add tests in `fee.rs` for SPL token transfer scenarios to ensure correct fee calculations.
>     - Add integration tests in `jupiter_integration.rs` for unknown tokens.
>   - **Misc**:
>     - Minor error handling improvements in `oracle.rs` and `auth.rs`.
>     - Update `encode_versioned_transaction` in `transaction.rs` to return `Result`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=solana-foundation%2Fkora&utm_source=github&utm_medium=referral)<sup> for 22d4f9218a3a2f7026e3b8fa0c4d1d27ef313715. You can [customize](https://app.ellipsis.dev/solana-foundation/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 83.7%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/18541937461)

### DIFF
--- a/crates/lib/src/oracle/jupiter.rs
+++ b/crates/lib/src/oracle/jupiter.rs
@@ -3,6 +3,7 @@ use crate::{
     constant::{JUPITER_API_LITE_URL, JUPITER_API_PRO_URL, SOL_MINT},
     error::KoraError,
     sanitize_error,
+    validator::math_validator,
 };
 use once_cell::sync::Lazy;
 use parking_lot::RwLock;
@@ -170,6 +171,8 @@ impl JupiterPriceOracle {
         let sol_price = jupiter_response
             .get(SOL_MINT)
             .ok_or_else(|| KoraError::RpcError("No SOL price data from Jupiter".to_string()))?;
+
+        math_validator::validate_division(sol_price.usd_price)?;
 
         // Convert all prices to SOL-denominated
         let mut result = HashMap::new();
--- a/crates/lib/src/token/token.rs
+++ b/crates/lib/src/token/token.rs
@@ -13,6 +13,7 @@ use crate::{
 };
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_sdk::{native_token::LAMPORTS_PER_SOL, pubkey::Pubkey};
+use spl_associated_token_account::get_associated_token_address_with_program_id;
 use std::{collections::HashMap, str::FromStr, time::Duration};
 
 #[cfg(not(test))]
@@ -188,9 +189,38 @@ impl TokenUtil {
                                     }
                                 }
                             }
-                            Err(_) => {
-                                // Skip if destination account doesn't exist or can't be fetched
-                                continue;
+                            Err(e) => {
+                                // If we get Account not found error, we try to match it to the ATA derivation for the fee payer
+                                // in case that ATA is being created in the current instruction
+                                if matches!(e, KoraError::AccountNotFound(_)) {
+                                    let spl_ata =
+                                        spl_associated_token_account::get_associated_token_address(
+                                            fee_payer,
+                                            mint_pubkey,
+                                        );
+                                    let token2022_ata =
+                                        get_associated_token_address_with_program_id(
+                                            fee_payer,
+                                            mint_pubkey,
+                                            &spl_token_2022::id(),
+                                        );
+
+                                    // If destination matches a valid ATA for fee payer, count as inflow
+                                    if *destination_address == spl_ata
+                                        || *destination_address == token2022_ata
+                                    {
+                                        mint_to_transfers
+                                            .entry(*mint_pubkey)
+                                            .or_default()
+                                            .push((*amount, false)); // inflow
+                                    }
+                                    // Otherwise, it's not fee payer's account, continue to next transfer
+                                } else {
+                                    // Skip if destination account doesn't exist or can't be fetched
+                                    // This could be problematic for non ATA token accounts created
+                                    // during the transaction
+                                    continue;
+                                }
                             }
                         }
                     }
@@ -272,7 +302,10 @@ impl TokenUtil {
         // Unpack the mint state with extensions
         let mint_state = token_program.unpack_mint(mint, &mint_data)?;
 
-        let mint_with_extensions = mint_state.as_any().downcast_ref::<Token2022Mint>().unwrap();
+        let mint_with_extensions =
+            mint_state.as_any().downcast_ref::<Token2022Mint>().ok_or_else(|| {
+                KoraError::SerializationError("Failed to downcast mint state.".to_string())
+            })?;
 
         // Check each extension type present on the mint
         for extension_type in mint_with_extensions.get_extension_types() {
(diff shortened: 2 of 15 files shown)


crates/lib/src/token/token.rs
```diff
@@ -140,6 +140,119 @@ impl TokenUtil {
...
+                    // Check if fee payer owns the destination (inflow)
+                    // We need to check the destination token account owner
+                    if let Some(mint_pubkey) = mint {
+                        // Get destination account to check owner
+                        match CacheUtil::get_account(rpc_client, destination_address, false).await {
+                            Ok(dest_account) => {
+                                let token_program =
+                                    TokenType::get_token_program_from_owner(&dest_account.owner)?;
+                                if let Ok(token_account) =
+                                    token_program.unpack_token_account(&dest_account.data)
```
amilz, line 182: Random thought, but i think T22 is backwards compatible here, so we could just always use T22 to unpack (we'd want to test to be sure, but pretty sure it just 0's out the extra data).

crates/lib/src/token/token.rs
```diff
@@ -140,6 +140,119 @@ impl TokenUtil {
...
+                                    if token_account.owner() == *fee_payer {
+                                        mint_to_transfers
+                                            .entry(*mint_pubkey)
+                                            .or_default()
+                                            .push((*amount, false)); // inflow
+                                    }
+                                }
+                            }
+                            Err(_) => {
+                                // Skip if destination account doesn't exist or can't be fetched
```
amilz: Is this okay? Suppose we have an ATA creation ix prior to this one--won't we skip b/c it's not found? maybe an alt to the `get_account` is to just derive atas and make sure it's valid ata

crates/lib/src/token/token.rs
```diff
@@ -140,6 +140,119 @@ impl TokenUtil {
...
+            3,
+            Duration::from_secs(1),
+            get_price_oracle(price_source.clone()),
+        );
+
+        let prices = oracle.get_token_prices(&mint_addresses).await?;
+
+        let mut mint_decimals = std::collections::HashMap::new();
+        for mint in mint_to_transfers.keys() {
+            let decimals = Self::get_mint_decimals(rpc_client, mint).await?;
```
amilz, line 249: this is maybe a longer term todo, but i wonder if we should have varying ttl for mint if we're just using to get owner/decimals...shouldn't ever really have to fetch fresh? wdyt

crates/lib/src/oracle/jupiter.rs
```diff
@@ -81,13 +81,29 @@ impl PriceOracle for JupiterPriceOracle {
...
+    ) -> Result<HashMap<String, TokenPrice>, KoraError> {
+        if mint_addresses.is_empty() {
+            return Ok(HashMap::new());
+        }
+
         // Try pro API first if API key is available, then fallback to free API
         if let Some(api_key) = &self.api_key {
             match self
-                .fetch_price_from_url(client, &self.pro_api_url, mint_address, Some(api_key))
+                .fetch_prices_from_url(client, &self.pro_api_url, mint_addresses, Some(api_key))
```
amilz, line 104: one thing we need to test is what happens if an unknown mint gets passed to jup? does it throw or just included some null data for that mint?

crates/lib/src/oracle/jupiter.rs
```diff
@@ -143,16 +166,33 @@ impl JupiterPriceOracle {
...
-        let price_data = jupiter_response
-            .get(mint_address)
-            .ok_or_else(|| KoraError::RpcError("No price data from Jupiter".to_string()))?;
 
-        let price = price_data.usd_price / sol_price.usd_price;
+        // Convert all prices to SOL-denominated
+        let mut result = HashMap::new();
+        for mint_address in mint_addresses {
+            if let Some(price_data) = jupiter_response.get(mint_address.as_str()) {
+                let price = price_data.usd_price / sol_price.usd_price;
```
amilz, line 181: does this need to be safe division?

crates/lib/src/token/token.rs
```diff
@@ -140,6 +140,119 @@ impl TokenUtil {
...
+            mint_decimals.insert(*mint, decimals);
+        }
+
+        // Calculate total value
+        let mut total_lamports = 0u64;
+
+        for (mint, transfers) in mint_to_transfers.iter() {
+            let price = prices
+                .get(&mint.to_string())
+                .ok_or_else(|| KoraError::RpcError(format!("No price data for mint {mint}")))?;
```
amilz, line 259: This could be annoying, but I think it's the right move...

tests/rpc/fee_estimation.rs
```diff
@@ -345,3 +349,82 @@ async fn test_estimate_fee_comprehensive_with_token_accounts_creation() {
         expected_minimum_fee + 50_000
     );
 }
+
+#[tokio::test]
+async fn test_estimate_fee_with_spl_token_transfer_from_fee_payer() {
```
amilz, line 354: I think since our functionality supports multiple...it's probably worth it to add a test for  handling muliple token transfer ix's. i think you could just do this w/ just 2 USDC ixns for testing. or adding a WSOL transfer

crates/lib/src/token/token.rs
dev-jodee: Yeah we can do that in the future, could use redis on disk like we do with the usage limit

crates/lib/src/oracle/jupiter.rs
dev-jodee: Crashes, added a test to validate

crates/lib/src/oracle/jupiter.rs
dev-jodee: f64 doesn't have checked, but I added some extra checks to make sure with divide by 0 (if somehow it would happen)

crates/lib/src/token/token.rs
dev-jodee: Yeah don't think continuing is safe to not have it sadly

crates/lib/src/token/token.rs
dev-jodee: I'll add deriving ATA but this won't wrok for non ATA token accounts, think I'll just error out here to fail safe

PR #300, opened 2026-01-16
PRO-606 feat: implement granular usage tracking with rule-based limits

dev-jodee:
## Summary
- Add flexible usage limit system with configurable rules for transactions, instructions, and bundles
- Implement UsageLimiter for enforcing limits with detailed violation reporting
- Enhance UsageTracker with metrics calculation and rule evaluation
- Integrate usage tracking with all RPC methods (sign, send, estimate, bundle operations)
- Add comprehensive test suite covering transaction limits, instruction limits, bundle limits, and multiple concurrent rules

## Test Plan
```bash
# Run new usage limit integration tests
just test-integration --filter usage_limit

# Run all integration tests
just test-integration

# Run unit tests
just test
```

The test suite validates:
- Transaction count limits (daily, hourly, per-request windows)
- Instruction count limits with window-based tracking
- Bundle limits for Jito bundle operations
- Multiple simultaneous rules with independent enforcement
- Detailed error responses with violation information

Closes PRO-606

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 82.0%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/21222481001)

### DIFF
--- a/tests/rpc/bundles.rs
+++ b/tests/rpc/bundles.rs
@@ -708,7 +708,13 @@ async fn test_sign_bundle_with_sign_only_indices_filter() {
     let response: serde_json::Value = ctx
         .rpc_call(
             "signBundle",
-            rpc_params![transactions, Option::<String>::None, false, sign_only_indices],
+            rpc_params![
+                transactions,
+                Option::<String>::None,
+                false,
+                Option::<String>::None,
+                sign_only_indices
+            ],
         )
         .await
         .expect("Failed to sign bundle");
@@ -766,7 +772,13 @@ async fn test_sign_bundle_out_of_bounds_sign_only_indices_fails() {
     let result: Result<serde_json::Value, _> = ctx
         .rpc_call(
             "signBundle",
-            rpc_params![transactions, Option::<String>::None, false, sign_only_indices],
+            rpc_params![
+                transactions,
+                Option::<String>::None,
+                false,
+                Option::<String>::None,
+                sign_only_indices
+            ],
         )
         .await;
 


tests/src/common/fixtures/auth-test.toml
```diff
@@ -98,8 +98,8 @@ allow_thaw_account = true
 [kora.usage_limit]
 enabled = false
 cache_url = "redis://redis:6379"
-max_transactions = 2
 fallback_if_unavailable = false
+rules = []
```
amilz, line 102: is this a required field now? if so, double check all toml files updated (e.g., `examples/`)

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -95,125 +82,303 @@ impl UsageTracker {
...
+            } = instruction
+            {
+                // Check if this is a payment to Kora by verifying the destination token account owner
+                // matches the payment destination
+                let destination_account =
+                    match CacheUtil::get_account(config, rpc_client, destination_address, true)
+                        .await
+                    {
+                        Ok(account) => account,
+                        Err(_) => continue, // Skip missing accounts
```
amilz: I think this is intended for if an account is not found in data or on network, continue--should we handle other types of errors and return the error in case something else happens?

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -95,125 +82,303 @@ impl UsageTracker {
...
+                return Err(KoraError::InternalServerError(
+                    "Usage limiter unavailable and fallback disabled".to_string(),
+                ));
+            }
+            return Ok(());
+        };
+
+        if tracker.has_instruction_rules() {
+            let _ = transaction.get_or_parse_system_instructions();
+            let _ = transaction.get_or_parse_spl_instructions();
```
amilz: do we need to propogate errors here? eg `transaction.get_or_parse_system_instructions()?; `

crates/lib/src/usage_limit/rules/instruction.rs
```diff
@@ -0,0 +1,363 @@
...
+                    let count = instructions
+                        .iter()
+                        .filter(|ix_data| {
+                            match ix_data {
+                                ParsedSystemInstructionData::SystemCreateAccount {
+                                    payer, ..
+                                } => {
+                                    // Count instructions where Kora IS the payer
+                                    // This tracks subsidized account creations
+                                    kora_signer == Some(*payer)
```
amilz, line 140: maybe i'm missing it but I don't see similar logic for ATA creation

amilz (commented):
Looks good. Couple of nits: 

Also: 

- One thing to note. here is that our limits are on signature requests, not executed transactions. This is _probably_ fine but might create some issues wrt to users doing testing/debugging (who might just call sign and throw away a txn -- using it in txn explorer or something like that). I can make sure we document this.

- Should we include `CreateAccountWithSeed` or `CreateATA` (non-idempotent)?

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -95,125 +82,303 @@ impl UsageTracker {
...
+                let _: Option<String> = conn.get("__usage_limiter_test__").await.map_err(|e| {
+                    KoraError::InternalServerError(format!(
+                        "Redis connection test failed: {}",
+                        sanitize_error!(e)
+                    ))
+                })?;
+
+                (Arc::new(RedisUsageStore::new(pool)), "Redis")
+            } else {
+                (Arc::new(InMemoryUsageStore::new()), "in-memory")
```
amilz, line 309: add something like (and maybe configure similar warning in CLI?)

```rust
      if !rules.is_empty() {                                                                                                                                      
          log::warn!(                                                                                                                                             
              "Usage limiting configured with in-memory store. \                                                                                                  
               Limits will NOT be shared across instances and will reset on restart. \                                                                            
               Configure 'cache_url' in [kora.usage_limit] for production deployments."                                                                           
          );                                                                                                                                                      
      }                                                                                                                                                           
```

tests/src/common/fixtures/auth-test.toml
dev-jodee: It's not, I added it to show it's available but it defaults to []

amilz (dismissed):

PR #301, opened 2026-01-17
feat(rpc): add reCAPTCHA support for protected endpoints

nasjuice:
This PR implements reCAPTCHA (v3) verification for protected RPC methods to prevent abuse.

### DIFF
--- a/crates/lib/src/rpc_server/auth.rs
+++ b/crates/lib/src/rpc_server/auth.rs
@@ -1,10 +1,7 @@
 use crate::{
-    constant::{X_API_KEY, X_HMAC_SIGNATURE, X_RECAPTCHA_TOKEN, X_TIMESTAMP},
-    rpc_server::{
-        middleware_utils::{
-            build_response_with_graceful_error, extract_parts_and_body_bytes, get_jsonrpc_method,
-        },
-        recaptcha_util::{RecaptchaConfig, RecaptchaValidated},
+    constant::{X_API_KEY, X_HMAC_SIGNATURE, X_TIMESTAMP},
+    rpc_server::middleware_utils::{
+        build_response_with_graceful_error, extract_parts_and_body_bytes, get_jsonrpc_method,
     },
 };
 use hmac::{Hmac, Mac};
@@ -16,17 +13,11 @@ use subtle::ConstantTimeEq;
 #[derive(Clone)]
 pub struct ApiKeyAuthLayer {
     api_key: String,
-    recaptcha_config: Option<RecaptchaConfig>,
 }
 
 impl ApiKeyAuthLayer {
     pub fn new(api_key: String) -> Self {
-        Self { api_key, recaptcha_config: None }
-    }
-
-    pub fn with_recaptcha(mut self, recaptcha_config: Option<RecaptchaConfig>) -> Self {
-        self.recaptcha_config = recaptcha_config;
-        self
+        Self { api_key }
     }
 }
 
@@ -34,17 +25,12 @@ impl ApiKeyAuthLayer {
 pub struct ApiKeyAuthService<S> {
     inner: S,
     api_key: String,
-    recaptcha_config: Option<RecaptchaConfig>,
 }
 
 impl<S> tower::Layer<S> for ApiKeyAuthLayer {
     type Service = ApiKeyAuthService<S>;
     fn layer(&self, inner: S) -> Self::Service {
-        ApiKeyAuthService {
-            inner,
-            api_key: self.api_key.clone(),
-            recaptcha_config: self.recaptcha_config.clone(),
-        }
+        ApiKeyAuthService { inner, api_key: self.api_key.clone() }
     }
 }
 
@@ -68,7 +54,6 @@ where
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
         let api_key = self.api_key.clone();
-        let recaptcha_config = self.recaptcha_config.clone();
         let mut inner = self.inner.clone();
 
         Box::pin(async move {
@@ -77,30 +62,20 @@ where
 
             let (parts, body_bytes) = extract_parts_and_body_bytes(request).await;
 
-            let method = get_jsonrpc_method(&body_bytes).unwrap_or_default();
-
             // Bypass auth for liveness endpoint
-            if method == "liveness" {
-                let new_body = Body::from(body_bytes);
-                let new_request = Request::from_parts(parts, new_body);
-                return inner.call(new_request).await;
+            if let Some(method) = get_jsonrpc_method(&body_bytes) {
+                if method == "liveness" {
+                    let new_body = Body::from(body_bytes);
+                    let new_request = Request::from_parts(parts, new_body);
+                    return inner.call(new_request).await;
+                }
             }
 
             // Check for API key header
-            let mut req = Request::from_parts(parts, Body::from(body_bytes.clone()));
+            let req = Request::from_parts(parts, Body::from(body_bytes));
             if let Some(provided_key) = req.headers().get(X_API_KEY) {
                 // Constant-time comparison prevents timing attacks
                 if provided_key.as_bytes().ct_eq(api_key.as_bytes()).into() {
-                    if let Some(config) = &recaptcha_config {
-                        if req.extensions().get::<RecaptchaValidated>().is_none() {
-                            let recaptcha_token =
-                                req.headers().get(X_RECAPTCHA_TOKEN).and_then(|v| v.to_str().ok());
-                            if let Err(resp) = config.validate(recaptcha_token, &method).await {
-                                return Ok(resp);
-                            }
-                            req.extensions_mut().insert(RecaptchaValidated);
-                        }
-                    }
                     return inner.call(req).await;
                 }
             }
@@ -110,23 +85,15 @@ where
     }
 }
 
-// ============ HMAC Authentication ============
-
 #[derive(Clone)]
 pub struct HmacAuthLayer {
     secret: String,
     max_timestamp_age: i64,
-    recaptcha_config: Option<RecaptchaConfig>,
 }
 
 impl HmacAuthLayer {
     pub fn new(secret: String, max_timestamp_age: i64) -> Self {
-        Self { secret, max_timestamp_age, recaptcha_config: None }
-    }
-
-    pub fn with_recaptcha(mut self, recaptcha_config: Option<RecaptchaConfig>) -> Self {
-        self.recaptcha_config = recaptcha_config;
-        self
+        Self { secret, max_timestamp_age }
     }
 }
 
@@ -138,7 +105,6 @@ impl<S> tower::Layer<S> for HmacAuthLayer {
             inner,
             secret: self.secret.clone(),
             max_timestamp_age: self.max_timestamp_age,
-            recaptcha_config: self.recaptcha_config.clone(),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


crates/lib/src/rpc_server/recaptcha.rs
```diff
@@ -0,0 +1,395 @@
...
+                _ => {
+                    return Ok(unauthorized_response);
+                }
+            };
+
+            let token = recaptcha_token.to_str().unwrap_or("");
+
+            match verify_recaptcha(&http_client, RECAPTCHA_VERIFY_URL, &secret, token).await {
+                Ok(score) => {
+                    if score < score_threshold {
```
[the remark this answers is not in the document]
nasjuice: It's the minimal acceptable score so values that are equal should also pass

crates/lib/src/constant.rs
```diff
@@ -6,14 +6,18 @@ pub const MIN_BALANCE_FOR_RENT_EXEMPTION: u64 = 2_039_280;
 pub const DEFAULT_INTEREST_MULTIPLIER: u128 = 100 * 24 * 60 * 60 / 10000 / (365 * 24 * 60 * 60);
 
 // HTTP Headers
+pub const X_RECAPTCHA_TOKEN: &str = "x-recaptcha-token";
 pub const X_API_KEY: &str = "x-api-key";
 pub const X_HMAC_SIGNATURE: &str = "x-hmac-signature";
 pub const X_TIMESTAMP: &str = "x-timestamp";
 pub const DEFAULT_MAX_TIMESTAMP_AGE: i64 = 300;
+pub const DEFAULT_RECAPTCHA_SCORE_THRESHOLD: f64 = 0.5;
+pub const DEFAULT_PROTECTED_METHODS: &[&str] = &["signTransaction", "signAndSendTransaction"];
```
dev-jodee: when you rebase on `release/2.2.0` make sure to add the *_bundle instruction as well

crates/lib/src/tests/config_mock.rs
```diff
@@ -430,7 +430,15 @@ impl Default for AuthConfigBuilder {
 
 impl AuthConfigBuilder {
     pub fn new() -> Self {
-        Self { config: AuthConfig { api_key: None, hmac_secret: None, max_timestamp_age: 10 } }
+        Self {
```
dev-jodee, line 448: Can you add to AuthConfigBuilder the with_recaptcha method as well

crates/lib/src/rpc_server/recaptcha.rs
```diff
@@ -0,0 +1,395 @@
...
+                        log::warn!(
+                            "reCAPTCHA verification failed: score {:.2} below threshold {:.2}",
+                            score,
+                            score_threshold
+                        );
+                        return Ok(unauthorized_response);
+                    }
+                }
+                Err(e) => {
+                    log::error!("reCAPTCHA verification error: {}", e);
```
dev-jodee: for error log always use sanitize_error! macro

crates/lib/src/config.rs
```diff
@@ -501,13 +505,22 @@ impl Default for UsageLimitConfig {
 pub struct AuthConfig {
     pub api_key: Option<String>,
     pub hmac_secret: Option<String>,
+    pub recaptcha_secret: Option<String>,
+    #[serde(default = "default_recaptcha_score_threshold")]
+    pub recaptcha_score_threshold: f64,
```
dev-jodee, line 544: this should be validated that it's within the valid bounds (error out in the config validator with the errors vec) when we validate

crates/lib/src/rpc_server/recaptcha.rs
```diff
@@ -0,0 +1,395 @@
...
+    >;
+
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
```
dev-jodee, line 54: Instead of having a completely new middleware, I'd be tempted to add a one liner in the auth.rs file instead 

RecaptchaUtil::validate_recaptcha() type of function that does all the logic for the validation and we only call the method if the config is enabled (quick bool check, and we save on duplicating extracting body, method name, etc.

Since it happens on all requests it adds up to a lot

crates/lib/src/rpc_server/recaptcha.rs
```diff
@@ -0,0 +1,395 @@
...
+async fn verify_recaptcha(
+    client: &Client,
+    verify_url: &str,
+    secret: &str,
+    token: &str,
+) -> Result<f64, KoraError> {
+    let response = client
+        .post(verify_url)
+        .form(&[("secret", secret), ("response", token)])
+        .send()
```
dev-jodee: we definitely want a timeout on this call

crates/lib/src/rpc_server/recaptcha.rs
```diff
@@ -0,0 +1,395 @@
...
+        .await
+        .map_err(|e| {
+            KoraError::RecaptchaError(format!("API call failed: {}", sanitize_error!(e)))
+        })?;
+
+    if !response.status().is_success() {
```
dev-jodee: logging here with the feature flag if sensitive logging you log all and if not you log sanitized

crates/lib/src/config.rs
```diff
@@ -501,13 +505,22 @@ impl Default for UsageLimitConfig {
 pub struct AuthConfig {
     pub api_key: Option<String>,
     pub hmac_secret: Option<String>,
+    pub recaptcha_secret: Option<String>,
```
dev-jodee, line 542: protected methods should be configurable in kora.toml

crates/lib/src/rpc_server/recaptcha.rs
nasjuice: I ended up adding the validate to the existing hmac/api services. Each layer has a `with_recaptcha` method that does a quick check and then makes the API calls necessary. 

This way we don't need an additional middleware

crates/lib/src/validator/config_validator.rs
```diff
@@ -281,6 +281,14 @@ impl ConfigValidator {
             }
         }
 
+        // Validate reCAPTCHA score threshold (must be between 0.0 and 1.0 inclusive)
+        let score_threshold = config.kora.auth.recaptcha_score_threshold;
+        if !(0.0..=1.0).contains(&score_threshold) {
```
dev-jodee: can we have constants here MIN / MAX recaptcha score

crates/lib/src/rpc_server/auth.rs
```diff
@@ -1,36 +1,173 @@
...
 use http::{Request, Response, StatusCode};
 use jsonrpsee::server::logger::Body;
+use reqwest::Client;
+use serde::Deserialize;
 use sha2::Sha256;
+use std::time::Duration;
 use subtle::ConstantTimeEq;
 
+#[derive(Debug, Deserialize)]
+struct RecaptchaVerifyResponse {
```
dev-jodee: Can we have a recaptcha util for separation / clarity?

crates/lib/src/rpc_server/auth.rs
```diff
@@ -1,36 +1,173 @@
...
+        )));
+    }
+
+    Ok(score)
+}
+
 #[derive(Clone)]
 pub struct ApiKeyAuthLayer {
     api_key: String,
+    recaptcha_config: RecaptchaConfig,
```
dev-jodee: I think logically, instead of having a recaptcha with a option secret, and then having to `unwrap` the secret every time in validate if required, we should just have the recaptcha config as optional

In all the calls for the auth middleware you just check is_some() and if is then you just call validate, the `secret` isnt an option anymore so you dont have to 

let secret = match &self.secret {
            Some(s) => s,
            None => return Ok(()),
        };

and I think its just cleaner in general

Also would allow you to not try to get the recaptcha token from the header if recaptcha is not validated (here)

`
        let recaptcha_token = request
                .headers()
                .get(X_RECAPTCHA_TOKEN)
                .and_then(|v| v.to_str().ok())
                .map(|s| s.to_string());
`

nasjuice: Good idea, I made the configs optional in the middlelayer

crates/lib/src/rpc_server/auth.rs
nasjuice: Moved everything related to recaptcha in utils file

crates/lib/src/rpc_server/server.rs
```diff
@@ -81,15 +93,19 @@ pub async fn run_rpc_server(rpc: KoraRpc, port: u16) -> Result<ServerHandles, an
         .layer(MethodValidationLayer::new(allowed_methods.clone()))
         // Add metrics collection layer
         .option_layer(metrics_layers.as_ref().and_then(|layers| layers.http_metrics_layer.clone()))
-        // Add authentication layer for API key if configured
+        // Add authentication layer for API key if configured (with integrated reCAPTCHA)
```
dev-jodee: Hmmm won't that duplicate the recaptcha calls if I have both API KEY and HMAC set up? Will run in both middlewares

crates/lib/src/rpc_server/recaptcha_util.rs
```diff
@@ -0,0 +1,147 @@
...
+            .send()
+            .await
+            .map_err(|e| {
+                KoraError::RecaptchaError(format!("API call failed: {}", sanitize_error!(e)))
+            })?;
+
+        if !response.status().is_success() {
+            let status = response.status();
+            #[cfg(feature = "unsafe-debug")]
+            log::error!("reCAPTCHA API returned error status: {}", status);
```
dev-jodee: Here you can actually print out the error since it's behind the unsafe debug flag

crates/lib/src/rpc_server/auth.rs
```diff
@@ -54,28 +68,44 @@ where
...
     fn call(&mut self, request: Request<Body>) -> Self::Future {
         let api_key = self.api_key.clone();
+        let recaptcha_config = self.recaptcha_config.clone();
         let mut inner = self.inner.clone();
 
         Box::pin(async move {
             let unauthorized_response =
                 build_response_with_graceful_error(None, StatusCode::UNAUTHORIZED, "");
 
+            let recaptcha_token = recaptcha_config.as_ref().and_then(|_| {
```
dev-jodee: This shoudl be in `if let Some(config) = &recaptcha_config {`

Only fetch the token if we have a config to check

crates/lib/src/rpc_server/auth.rs
```diff
@@ -146,15 +187,24 @@ where
             let signature_header = request.headers().get(X_HMAC_SIGNATURE).cloned();
             let timestamp_header = request.headers().get(X_TIMESTAMP).cloned();
 
+            let recaptcha_token = recaptcha_config.as_ref().and_then(|_| {
```
dev-jodee: Same as above

crates/lib/src/rpc_server/server.rs
nasjuice: Added a validatedrecaptcha extensions to not run in both

crates/lib/src/rpc_server/recaptcha_util.rs
nasjuice: I think ill just consolidate it by using sanitize_error since it checks for cfg

PR #302, opened 2026-01-20
feat: add transactions_to_sign parameter to bundle endpoints

dev-jodee:
## Summary
- Add optional `transactions_to_sign` parameter to `signBundle` and `signAndSendBundle` endpoints
- Enable partial bundle signing where clients can specify which transactions Kora should sign
- Implement validation for out-of-bounds transaction indices
- Fix bundle integration test parameter format to use positional params correctly

## Test Plan
All 20 bundle integration tests passing:
- Partial signing tests verify correct transaction selection and merging
- Out-of-bounds validation test confirms proper error handling
- Existing bundle tests confirm backward compatibility (all transactions signed by default)

Run tests with:
```bash
just test-integration --filter regular
```

Closes PRO-743

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 82.1%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/21215391828)

### DIFF
--- a/crates/lib/src/bundle/helper.rs
+++ b/crates/lib/src/bundle/helper.rs
@@ -13,7 +13,7 @@ use crate::{
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_commitment_config::CommitmentConfig;
 use solana_sdk::{instruction::Instruction, pubkey::Pubkey};
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 pub struct BundleProcessor {
     pub resolved_transactions: Vec<VersionedTransactionResolved>,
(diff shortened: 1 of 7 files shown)


crates/lib/src/bundle/helper.rs
```diff
@@ -23,6 +23,61 @@ pub struct BundleProcessor {
...
+
+        Ok((filtered, indices.to_vec()))
+    }
+
+    /// Merge signed transactions back into the original list, preserving order.
+    /// Transactions at indices in `signed_indices` are replaced with corresponding signed versions.
+    pub fn merge_signed_transactions(
+        original_transactions: &[String],
+        signed_transactions: Vec<String>,
+        signed_indices: &[usize],
```
amilz: maybe add a validation here to throw if `signed_transactions.len() != signed_indices.len() `

crates/lib/src/bundle/helper.rs
```diff
@@ -23,6 +23,61 @@ pub struct BundleProcessor {
...
 
 impl BundleProcessor {
+    /// Extract transactions at specified indices for processing.
+    /// Returns (filtered_transactions, indices_used).
+    /// If `transactions_to_sign` is None, returns all transactions with all indices.
+    pub fn extract_transactions_to_process(
+        transactions: &[String],
+        transactions_to_sign: Option<&[usize]>,
+    ) -> Result<(Vec<String>, Vec<usize>), KoraError> {
+        let indices: &[usize];
```
amilz: should we add a hashset validation to ensure no dupes?

crates/lib/src/bundle/helper.rs
```diff
@@ -23,6 +23,61 @@ pub struct BundleProcessor {
...
+        signed_transactions: Vec<String>,
+        signed_indices: &[usize],
+    ) -> Vec<String> {
+        let mut signed_iter = signed_transactions.into_iter();
+        let signed_set: std::collections::HashSet<usize> = signed_indices.iter().copied().collect();
+
+        (0..original_transactions.len())
+            .map(|idx| {
+                if signed_set.contains(&idx) {
+                    signed_iter.next().unwrap()
```
amilz: Check me on this, but I think if a user passes indices in descending order (e.g., `[2,0]`) this risks reordering their transactions?

tests/rpc/bundles.rs
```diff
@@ -665,3 +665,115 @@ async fn test_sign_bundle_insufficient_payment_error() {
...
+            .with_transfer(&sender.pubkey(), &recipient, 10 + i)
+            .build()
+            .await
+            .expect("Failed to create transaction");
+        transactions.push(tx);
+    }
+
+    // Process only transactions at indices 0 and 2
+    // Pass transactions_to_sign as positional params: [transactions, signer_key, sig_verify, transactions_to_sign]
+    let transactions_to_sign: Vec<usize> = vec![0, 2];
```
amilz: should we add a test with an empt vec? I assume we expect all signed?

crates/lib/src/bundle/helper.rs
dev-jodee: No need will use hashmap for duplicate

crates/lib/src/bundle/helper.rs
dev-jodee: hashmap fix for that as well :)

tests/rpc/bundles.rs
dev-jodee: Empty vec would sign no transactions, its an optional parameter, you pass none if you dont want to use it, if you explicitely pass an empty vec then it'll just forward it to Jito. 

Dont think we should handle empty vec as none, think it'll bring in more confusion and it's common that you just pass null if you don't want to use the parameter

PR #315, opened 2026-01-28
feat(lighthouse): add fee payer protection for sign endpoints

dev-jodee:
## Summary
- Add lighthouse balance assertion to protect fee payer from drainage attacks
- Implement `will_send` parameter to control lighthouse application: `signTransaction`/`signBundle` add lighthouse (client re-signs), `signAndSendTransaction`/`signAndSendBundle` skip lighthouse (preserve client signatures)
- Add config validation for lighthouse settings with `fail_if_transaction_size_overflow` option
- Include 5 integration tests covering legacy/V0 transactions, bundles, and end-to-end client re-signing flow

## Test Plan
```bash
just test-integration --filter lighthouse
```

All 5 tests pass:
- `test_sign_transaction_with_lighthouse_legacy` - Verify lighthouse assertion added to legacy transactions
- `test_sign_transaction_with_lighthouse_v0` - Verify lighthouse assertion added to V0 transactions
- `test_sign_bundle_with_lighthouse` - Verify lighthouse assertion added to last transaction in bundle
- `test_lighthouse_end_to_end_with_client_resign` - Full flow: sign → Kora adds lighthouse → client re-signs → simulate succeeds
- `test_lighthouse_end_to_end_v0_with_client_resign` - Same flow for V0 transactions

Closes PRO-151

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 81.7%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/21482162653)

### DIFF
--- a/crates/lib/src/bundle/helper.rs
+++ b/crates/lib/src/bundle/helper.rs
@@ -227,15 +227,15 @@ impl BundleProcessor {
                 );
             }
 
-            // Add lighthouse assertion to last transaction only when NOT sending (client will re-sign)
-            // Skip for sign-and-send flows as modifying the message would invalidate client signatures
-            if !will_send && i == tx_count - 1 {
+            // Add lighthouse assertion only to last transaction in bundle
+            if i == tx_count - 1 {
                 LighthouseUtil::add_fee_payer_assertion(
                     &mut resolved.transaction,
                     rpc_client,
                     fee_payer,
                     self.total_solana_estimated_fee,
                     &config.kora.lighthouse,
+                    will_send,
                 )
                 .await?;
             }
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -32,12 +32,12 @@ pub struct Config {
 }
 
 #[derive(Clone, Serialize, Deserialize, ToSchema)]
+#[serde(default)]
 pub struct MetricsConfig {
     pub enabled: bool,
     pub endpoint: String,
     pub port: u16,
     pub scrape_interval: u64,
-    #[serde(default)]
     pub fee_payer_balance: FeePayerBalanceMetricsConfig,
 }
 
@@ -146,16 +146,15 @@ impl ValidationConfig {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema, Default)]
+#[serde(default)]
 pub struct FeePayerPolicy {
-    #[serde(default)]
     pub system: SystemInstructionPolicy,
-    #[serde(default)]
     pub spl_token: SplTokenInstructionPolicy,
-    #[serde(default)]
     pub token_2022: Token2022InstructionPolicy,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema, Default)]
+#[serde(default)]
 pub struct SystemInstructionPolicy {
     /// Allow fee payer to be the sender in System Transfer/TransferWithSeed instructions
     pub allow_transfer: bool,
@@ -166,7 +165,6 @@ pub struct SystemInstructionPolicy {
     /// Allow fee payer to be the account in System Allocate/AllocateWithSeed instructions
     pub allow_allocate: bool,
     /// Nested policy for nonce account operations
-    #[serde(default)]
     pub nonce: NonceInstructionPolicy,
 }
 
@@ -323,6 +321,7 @@ impl Token2022Config {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(default)]
 pub struct EnabledMethods {
     pub liveness: bool,
     pub estimate_transaction_fee: bool,
@@ -335,11 +334,8 @@ pub struct EnabledMethods {
     pub get_config: bool,
     pub get_version: bool,
     /// Bundle methods (require bundle.enabled = true)
-    #[serde(default)]
     pub estimate_bundle_fee: bool,
-    #[serde(default)]
     pub sign_and_send_bundle: bool,
-    #[serde(default)]
     pub sign_bundle: bool,
 }
 
@@ -454,18 +450,6 @@ impl Default for EnabledMethods {
     }
 }
 
-fn default_max_timestamp_age() -> i64 {
-    DEFAULT_MAX_TIMESTAMP_AGE
-}
-
-fn default_max_request_body_size() -> usize {
-    DEFAULT_MAX_REQUEST_BODY_SIZE
-}
-
-fn default_recaptcha_score_threshold() -> f64 {
-    DEFAULT_RECAPTCHA_SCORE_THRESHOLD
-}
-
 #[derive(Clone, Serialize, Deserialize, ToSchema)]
 pub struct CacheConfig {
     /// Redis URL for caching (e.g., "redis://localhost:6379")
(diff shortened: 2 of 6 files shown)


crates/lib/src/config.rs
```diff
@@ -535,6 +539,28 @@ pub struct BundleConfig {
...
 
+/// Configuration for Lighthouse assertions to protect fee payer balance
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct LighthouseConfig {
+    /// Enable Lighthouse assertions for fee payer protection
+    #[serde(default)]
+    pub enabled: bool,
+    /// If true, fail when adding assertion would exceed transaction size limit.
+    /// If false, skip adding the assertion silently.
+    #[serde(default = "default_fail_on_overflow")]
```
amilz: is this needed since you impl Default below?

crates/lib/src/rpc_server/method/sign_transaction.rs
```diff
@@ -62,7 +62,7 @@ pub async fn sign_transaction(
     .await?;
 
     let (signed_transaction, _) =
-        resolved_transaction.sign_transaction(config, &signer, rpc_client).await?;
+        resolved_transaction.sign_transaction(config, &signer, rpc_client, false).await?;
```
amilz, line 65: does `sign_and_send` need to add a `true` as well?  and do we need a test somewhere to catch that sign and send shoudln't be adding the ix?

crates/lib/src/lighthouse/assertion.rs
```diff
@@ -0,0 +1,344 @@
...
+                    "Adding Lighthouse assertion would exceed transaction size limit ({} > {})",
+                    new_size, MAX_TRANSACTION_SIZE
+                )));
+            } else {
+                log::warn!(
+                    "Lighthouse assertion would exceed transaction size limit ({} > {}). Skipping.",
+                    new_size,
+                    MAX_TRANSACTION_SIZE
+                );
+                return Ok(());
```
amilz, line 199: so the end result here is a signed transaction returned to the user _WITHOUT_ an assertion? if i'm not reviewing the server logs...the client might assume that an assertion is included? I get that `fail_if_transaction_size_overflow` is basically a risk the server operator is willing to take, but i'm wondering if we should include a bool in the response data `appended_lighthouse`. I lean no...b/c it's a server config but flagging for consideration. WDYT? /

crates/lib/src/lighthouse/assertion.rs
```diff
@@ -0,0 +1,344 @@
...
+        fee_payer: &Pubkey,
+        estimated_fee: u64,
+        config: &LighthouseConfig,
+    ) -> Result<(), KoraError> {
+        if !config.enabled {
+            return Ok(());
+        }
+
+        let current_balance = rpc_client.get_balance(fee_payer).await?;
+        let min_expected = current_balance.saturating_sub(estimated_fee);
```
amilz, line 55: 1) is there an edge case we should catch were balance - fee <= 0?

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -302,6 +305,19 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
         let estimated_fee = TransactionFeeUtil::get_estimate_fee_resolved(rpc_client, self).await?;
         validator.validate_lamport_fee(estimated_fee)?;
 
+        // Add lighthouse assertion only when NOT sending (client will re-sign)
+        // Skip for sign-and-send flows as modifying the message would invalidate client signatures
+        if !will_send {
```
amilz: nit/preference, imo `if config.kora.lighthouse & !will_send`. just reads a little cleaner to me.

tests/lighthouse/lighthouse_tests.rs
amilz, line 1: should we make sure here that `_and_send` methods don't have the assertion. 
also should some of our non-lighthouse tests include an assertion there's no added LH ix?

crates/lib/src/validator/config_validator.rs
```diff
@@ -341,6 +341,23 @@ impl ConfigValidator {
             }
         }
 
+        // Validate lighthouse configuration
+        if config.kora.lighthouse.enabled {
```
amilz, line 345: I think we should also add-- a warning if enabled AND if `_and_send` are enabled that those methods will get sent to the network _without_ the LH assertions

tests/src/common/fixtures/test-programs/lighthouse.so
amilz, line 1: should we set up a dependabot or something to make sure we don't miss any bumps?

crates/lib/src/bundle/helper.rs
```diff
@@ -223,6 +227,19 @@ impl BundleProcessor {
                 );
             }
 
+            // Add lighthouse assertion to last transaction only when NOT sending (client will re-sign)
+            // Skip for sign-and-send flows as modifying the message would invalidate client signatures
+            if !will_send && i == tx_count - 1 {
```
amilz: same as comment in single txn...imo `if config && !will_send &&...` then add

amilz (commented):
per our chat, IMO we probably need to check for a CU ix and add CUs if we're appending  (or something like that). dropped a few smaller comments as well.

crates/lib/src/lighthouse/assertion.rs
```diff
@@ -0,0 +1,344 @@
...
+        rpc_client: &RpcClient,
+        fee_payer: &Pubkey,
+        estimated_fee: u64,
+        config: &LighthouseConfig,
+    ) -> Result<(), KoraError> {
+        if !config.enabled {
+            return Ok(());
+        }
+
+        let current_balance = rpc_client.get_balance(fee_payer).await?;
```
amilz: nit but maybe add error context `Failed to fetch fee payer balance in Lighthouse Assertion` or something

crates/lib/src/rpc_server/method/sign_transaction.rs
dev-jodee: Already there in VersionedTransaction

```
async fn sign_and_send_transaction(
        &mut self,
        config: &Config,
        signer: &std::sync::Arc<Signer>,
        rpc_client: &RpcClient,
    ) -> Result<(String, String), KoraError> {
        // Payment validation is handled in sign_transaction
        let (transaction, encoded) =
            self.sign_transaction(config, signer, rpc_client, true).await?;

        // Send and confirm transaction
        let signature = rpc_client
            .send_and_confirm_transaction(&transaction)
            .await
            .map_err(|e| KoraError::RpcError(e.to_string()))?;

        Ok((signature.to_string(), encoded))
    }
```

crates/lib/src/config.rs
dev-jodee: serde default on prop needed that but re-orged it in a better way 👍

crates/lib/src/lighthouse/assertion.rs
dev-jodee: Yeah think since it's a server configuration, it's a risk the operator wants to take just like a lot of our other configs where we "allow" it but it could cause strange / risky signing. Philosophy is to never block an operator from doing something, help them make the best decision / have the controls they want, but in the end it's a decision on their side base on risk tolerance

crates/lib/src/lighthouse/assertion.rs
dev-jodee: I'll log something, but if that's the case, min expected with sat sub = 0, txn would fail elsewhere if fee payer doesn't have the balance for all the fees

crates/lib/src/transaction/versioned_transaction.rs
dev-jodee: added it inside the call instead, I prefer keeping caller clean and add if logic in callee

tests/src/common/fixtures/test-programs/lighthouse.so
dev-jodee: Hasn't updated in 9 months https://github.com/Jac0xb/lighthouse, wdyt?

amilz: prob fine

PR #352, opened 2026-02-24
fix(cli): validate pubkey format for allowed_programs in config

raushan728:
Adds missing base58 format validation for `allowed_programs` in the configuration.

While tokens and accounts were already being validated via `TokenUtil`, `allowed_programs` was missing this format check. This ensures `config validate` fails fast with a clear error if a user accidentally types an invalid program address in `kora.toml`.

### DIFF
--- a/crates/lib/src/validator/config_validator.rs
+++ b/crates/lib/src/validator/config_validator.rs
@@ -385,6 +385,16 @@ impl ConfigValidator {
             }
         }
 
+        // Validate base58 pubkey format for allowed_programs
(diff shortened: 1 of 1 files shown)


raushan728:
hi @amilz and @dev-jodee,

It looks like the `Integration Tests (regular)` CI job failed due to an external issue with the Jupiter API

could you guys take a look or re-run the job when you get a chance? thanks!

dev-jodee (approved):

PR #354, opened 2026-02-25
feat(sdk): implement paymaster client with kit plugin interfaces

amilz:
## Summary

- Add `createDefaultKoraClient()` — a high-level Kit client satisfying `ClientWithPayer`, `ClientWithRpc`, `ClientWithTransactionPlanning`, and `ClientWithTransactionSending`, enabling composability with Kit program plugins like `tokenProgram()`
- Uses Kit's plugin composition: rpc, payer, planner, executor wired via `planAndSendTransactions()`
- `rpcUrl` is required — CU estimation is always on (simulation-based) and program plugins work out of the box
- Placeholder payment instruction matching uses `parseTransferInstruction` from `@solana-program/token`
- Token-2022 support via `tokenProgramId`

### Review feedback (mcintyre94)

- Evaluated `getTransferToATAInstructionPlanAsync` but kept manual approach: ATA creation is operator responsibility, decimals validation is redundant (Kora handles fees)
- Created [PRO-983](https://linear.app/solana-fndn/issue/PRO-983) for future server-side ATA cost deduction

## Usage

```typescript
import { createDefaultKoraClient } from '@solana/kora';
import { tokenProgram } from '@solana-program/token';
import { address } from '@solana/kit';

const defaultKoraClient = await createDefaultKoraClient({
  endpoint: 'https://kora.example.com',
  rpcUrl: 'https://api.mainnet-beta.solana.com',
  feeToken: address('EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v'),
  feePayerWallet: myWalletSigner,
});

// Compose with Kit program plugins
const client = defaultKoraClient.use(tokenProgram());

// Transfer USDC via the token plugin — Kora handles gas fees transparently
await client.token.instructions
  .transferToATA({
    authority: myWalletSigner,
    recipient: address('...'),
    mint: address('EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v'),
    amount: 1_000_000,
    decimals: 6,
  })
  .sendTransaction();
```

## Test Plan

- `pnpm run build` / `pnpm run type-check` — clean
- `pnpm test kit-client.test.ts` — 18/18 pass
- `pnpm test unit.test.ts` — 28/28 pass
- `make integration-test` — all phases pass

### Integration tests added

- **Planner**: placeholder present with `amount=0n`, user instructions packed correctly
- **Executor**: placeholder resolved with real fee, multiple instructions handled
- **Plugin composition**: `tokenProgram().transferToATA().sendTransaction()` end-to-end
- **Free pricing** (`typescript_free` phase): fee=0 placeholder stripping verified

Closes PRO-934

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 81.0%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/22774382135)

### DIFF
--- a/.github/workflows/typescript-integration.yml
+++ b/.github/workflows/typescript-integration.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test-group: [typescript_basic, typescript_auth]
+        test-group: [typescript_basic, typescript_auth, typescript_free]
     steps:
       - uses: actions/checkout@v4
 
--- a/tests/src/common/fixtures/kora-free-test.toml
+++ b/tests/src/common/fixtures/kora-free-test.toml
@@ -32,6 +32,7 @@ allowed_programs = [
     "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb",  # Token-2022 Program
     "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL", # Associated Token Program
     "AddressLookupTab1e1111111111111111111111111",  # Address Lookup Table Program
+    "ComputeBudget111111111111111111111111111111",  # Compute Budget Program
     "Bcdikjss8HWzKEuj6gEQoFq9TCnGnk6v3kUnRU1gb6hA", # Custom Transfer Hook Program Example
 ]
 allowed_tokens = [


sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+        config?: { abortSignal?: AbortSignal },
+    ) => Promise<SuccessfulSingleTransactionPlanResult>;
+    /**
+     * Send one or more transactions.
+     * Accepts instructions, instruction plans, transaction messages, or transaction plans.
+     */
+    sendTransactions: (
+        input: InstructionPlanInput | TransactionPlanInput,
+        config?: { abortSignal?: AbortSignal },
+    ) => Promise<TransactionPlanResult>;
```
mcintyre94: I think you can simplify this type to something like:

```ts
Client<{
  kora: KoraApi,
 paymentAddress: Address;
 // all your custom fields
} & ClientWithPayer & ClientWithTransactionPlanning & ClientWithTransactionSending
```
  
See https://github.com/anza-xyz/kit/tree/main/packages/plugin-interfaces + https://github.com/anza-xyz/kit/blob/main/packages/plugin-core/src/client.ts

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+ */
+export interface KoraPaymasterClient {
+    /** Access to the raw Kora RPC methods (Kit-typed) */
+    readonly kora: KoraApi;
+    /** The Kora payer NoopSigner — satisfies ClientWithPayer */
+    readonly payer: TransactionSigner;
+    /** Alias for payer */
+    readonly payerSigner: TransactionSigner;
+    /** The Kora payer address */
+    readonly payerAddress: Address;
```
mcintyre94: Guessing these are required for backward compatibility?

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+        input: InstructionPlanInput | TransactionPlanInput,
+        config?: { abortSignal?: AbortSignal },
+    ) => Promise<TransactionPlanResult>;
+}
+
+// Compile-time conformance: ensure KoraPaymasterClient satisfies Kit plugin interfaces.
+// If Kit changes these interfaces, this will produce a type error at build time.
+const _assertPayer: KoraPaymasterClient extends ClientWithPayer ? true : never = true;
+const _assertPlanning: KoraPaymasterClient extends ClientWithTransactionPlanning ? true : never = true;
+const _assertSending: KoraPaymasterClient extends ClientWithTransactionSending ? true : never = true;
```
mcintyre94: You'd get these checks for free if you changed `KoraPaymasterClient` to just extend these types

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+        tokenProgram: tokenProgram,
+        mint: config.feeToken,
+    });
+
+    const paymentIx: Instruction = getTransferInstruction({
+        source: sourceTokenAccount,
+        destination: destinationTokenAccount,
+        authority: feePayerWalletSigner,
+        amount: fee_in_token,
+    });
```
mcintyre94: Nit: if you want to simplify this you could use `getTransferToATAInstructionPlanAsync` which does both  `findAssociatedTokenPda` for you: https://github.com/solana-program/token/blob/e7fe2e31c1b1f6c437ba3bb8911b687976d22e87/clients/js/src/transferToATA.ts#L80

You'd be returning an `InstructionPlan` (it does a create idempotent on the recipient side + transfer checked), but will be compatible if you're using a plugin to send this

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+                config,
+                paymentAddr,
+                config.feePayerWallet,
+                prePaymentTx,
+            );
+
+            let finalTx: Base64EncodedWireTransaction;
+            if (paymentIx !== null) {
+                // Rebuild message with payment IX appended
+                const msgWithPayment = appendTransactionMessageInstructions([paymentIx], msgForEstimation);
```
mcintyre94: If `paymentIx` were an `InstructionPlan` (my above comment), you could use `appendTransactionMessageInstructionPlan` here.

That said - you should try to avoid adding new instructions here, and use the same pattern that we use for compute budget. Add a placeholder in the planner, and then update that instruction in the executor.

The reason for this is that we have an API that lets you plan transactions from an array of instructions/instruction plans. When you use this, we pack into transactions for you, using the transaction in the planner as the starting point. But this only works if that message in the planner is aware of all the instructions that will be required, in addition to the ones it's packing.

With your current implementation, if someone used the planning API with enough instructions to require multiple transactions, you could end up over the transaction size limit after appending the additional instruction to a fully packed transaction here.

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+            if (result.signature) {
+                return signature(result.signature);
+            }
+            const signedTxBytes = getBase64Encoder().encode(result.signed_transaction);
+            const decodedTx = getTransactionDecoder().decode(signedTxBytes);
+            return getSignatureFromTransaction(decodedTx);
+        },
+    });
+
+    // --- Build the 4 interface methods ---
```
mcintyre94: I think you could just use our `planAndSendTransactions()` plugin here: https://github.com/anza-xyz/kit-plugins/blob/1ab8f82316e05d5d91df61b9b38fce519e75868c/packages/kit-plugin-instruction-plan/src/core.ts#L91

It just adds these 4 methods to any client that already has a planner + executor

sdks/ts/src/paymaster.ts
```diff
@@ -0,0 +1,389 @@
...
+        use<T extends object>(plugin: (client: KoraPaymasterClient) => T): T & KoraPaymasterClient {
+            const additions = plugin(client);
+            return { ...client, ...additions } as T & KoraPaymasterClient;
+        },
+
+        planTransaction,
+        planTransactions,
+        sendTransaction,
+        sendTransactions,
+    };
```
mcintyre94: The way to compose a client is with plugins. You'd make a plugin that adds planner + executor, that looks like eg: https://github.com/anza-xyz/kit-plugins/blob/1ab8f82316e05d5d91df61b9b38fce519e75868c/packages/kit-plugin-rpc/src/transaction-planner.ts#L40

And then you'd write your client like this:

```
return createEmptyClient()
  .use(payer(signerPayer)) // our payer plugin
  .use(koraPayer()) // make a plugin that sets payerSigner, payerAddress based on `client.payer`
  .use(koraPaymentAddress(paymentAddr)) // set paymentAddress (just a fixed field)
  .use(koraTransactionPlannerAndExecutor(...)) // set the planner + executor
  .use(planAndSendTransactions()) // use our plugin to generate all the sendTransaction etc APIs for you using your planner + executor
```

sdks/ts/src/paymaster.ts
amilz: Though I like this suggestion, I'm going to skip it for now. 
1. the planner `getTransferToATAInstructionPlanAsync` always adds `getCreateAssociatedTokenIdempotentInstruction`. And in our case, we're expecting the Kora node operator to have an ATA always and though we could force the node to pay the ata creation, the way kora works under the hood rn is that the client would end up having to pay for this (i'll make a ticket to see if we can prevent that)
2. This uses `getTransferCheckedInstruction` under the hood. This is _okay_ but adds the burden of decimal management on the client and doesn't really add any value here. Kora does the decimal math under the hood and provides the payment amount via our estimate method.

sdks/ts/src/paymaster.ts
amilz: Got it. 
I may need to pick your brain about our release branch which includes jito bundles, but I'll save that for another day. 

https://github.com/solana-foundation/kora/blob/f8d27f8c9f091dfb965e83d6677b1d57a0563158/sdks/ts/src/client.ts#L325

dev-jodee (approved):

PR #393, opened 2026-03-18
feat(fee): deduct operator ATA rent from client fee via dynamic_ata_deduction config

dev-jodee:
Fixes #377

Recreated from #382 (originally opened against `release/2.2.0`, which was merged into `main`, automatically closing the PR). All commits preserve original authorship by @raushan728 (Raushan Kumar).

---

Adds `dynamic_ata_deduction` config flag (default: `false`). When enabled, Kora detects `createAssociatedTokenIdempotent` instructions targeting the operator's payment address for the fee payment token and deducts the actual ATA rent (verified via matching `SystemCreateAccount`) from the client's fee outflow.

- Deduction only applies when a matching `SystemCreateAccount` exists (no deduction if ATA already exists — idempotent no-op case)
- Scoped to the specific mint used for fee payment, not any operator ATA
- Deduplicated per `(wallet_owner, mint)` pair
- Fee payer verified as ATA creation payer

Existing flow is completely unaffected when flag is disabled.

---

## Additional fixes on top of #382 (security hardening)

- **Token-2022 rent fix**: Replaced hardcoded `MIN_BALANCE_FOR_RENT_EXEMPTION` lamport equality check with actual lamports from the matched `SystemCreateAccount`. Token-2022 ATAs have the `ImmutableOwner` extension (170 bytes vs 165) so their rent is higher — the hardcoded check always returned `false` for them, silently disabling deduction.

- **SystemCreateAccount payer verification**: The deduction now verifies the matching `SystemCreateAccount` was paid by `fee_payer_pubkey`. Without this, if another account paid for the ATA creation, the rent was never added to the outflow but the deduction still subtracted it, producing an under-count of actual drain.

- **Non-allowed token allowlist gate**: `payment_mints` is now filtered by `config.validation.supports_token()`. Previously any token transferred to the operator's ATA (including non-allowed ones) could trigger the deduction, allowing a client to reduce the `max_allowed_lamports` drain-protection check by ATA rent per non-allowed token.

- **PDA derivation check in `parse_ata_instructions`**: Added the same PDA derivation verification to the parser used by the fee.rs path, consistent with the fix in `find_ata_creation_for_destination` (token.rs). Rejects crafted instructions with mismatched `wallet_owner`/`mint`/`ata_address` values.

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 83.9%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/23297551234)


scutuatua-crypto (approved):

crates/lib/src/config.rs
```diff
@@ -137,6 +137,13 @@ pub struct ValidationConfig {
     /// When >0, prices with a block_id older than `current_slot - max_price_staleness_slots` are rejected.
     #[serde(default)]
     pub max_price_staleness_slots: u64,
+    /// When enabled, deducts ATA creation rent from client fee
+    /// if operator payment address ATA does not exist.
+    /// Default: false.
+    /// WARNING: Enabling with dynamic pricing may allow clients to satisfy fees
+    /// using rent lamports whose SOL value exceeds the required token payment.
```
amilz, line 144: this is true--but it should be limited to ATA rent. worth adding an assertion and test somewhere that the delta doesn't exceed that rent amount?

crates/lib/src/token/token.rs
amilz, line 1: wdyt about adding a test `find_ata_creation_for_destination` with PDA mismatch— 
Should construct an ATA instruction where `ata_address == destination_address` but `wallet_owner`/`mint` derive to a *different* address, and assert the function returns `None`.

crates/lib/src/transaction/instruction_util.rs
amilz, line 1: wdyt adding `parse_ata_instructions` with PDA mismatch. 
Should verify that a malformed instruction with mismatched PDA gets skipped.

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -3603,4 +3683,75 @@ mod tests {
...
+                &wallet_owner,
+                &mint,
+            );
+
+        // 1. Test with ATA instruction
+        let instruction = create_associated_token_account_idempotent(
+            &payer.pubkey(),
+            &wallet_owner,
+            &mint,
+            &spl_token_interface::ID,
```
amilz, line 3719: probably worth adding a t22 test

crates/lib/src/fee/fee.rs
```diff
@@ -466,11 +467,166 @@ impl FeeConfigUtil {
...
+                    );
+                }
+            }
+        }
 
         Ok(total)
     }
 }
 
+async fn collect_payment_mints(
```
amilz: want to add any unit tests?

crates/lib/src/fee/fee.rs
```diff
@@ -466,11 +467,166 @@ impl FeeConfigUtil {
...
+                        mint_pubkey,
+                        &program_id,
+                    );
+                    if *destination_address == operator_ata {
+                        payment_mints.insert(*mint_pubkey);
+                    }
+                } else {
+                    // For regular transfers, check the destination account owner
+                    if let Ok(dest_account) =
+                        CacheUtil::get_account(config, rpc_client, destination_address, false).await
```
amilz: i think if this errors and we miss a mint it could cause estimate math to throw--prob worth warning or debug on err?

crates/lib/src/fee/fee.rs
```diff
@@ -466,11 +467,166 @@ impl FeeConfigUtil {
...
+                            TokenType::get_token_program_from_owner(&dest_account.owner)
+                        {
+                            if let Ok(token_account) =
+                                token_program.unpack_token_account(&dest_account.data)
+                            {
+                                let token_mint = token_account.mint();
+                                if token_account.owner() == *payment_destination
+                                    && config.validation.supports_token(&token_mint.to_string())
+                                {
+                                    payment_mints.insert(token_mint);
```
amilz: should we do ata check here? 

```rs
  let expected_ata = get_associated_token_address_with_program_id(
      payment_destination,
      &token_mint,
      &dest_account.owner,
  );
  if *destination_address == expected_ata {
      payment_mints.insert(token_mint);
  }
```

amilz (dismissed):
let's chat!

PR #424, opened 2026-04-09
feat: structured numeric RPC error codes

raushan728:
Replaces string-based error parsing with stable numeric codes (-32000 to -32099).

- Added `KoraErrorCode` enum in Rust and TS
- Added `KoraError` class in TS SDK clients can now use `instanceof KoraError` and switch on `error.code`

Breaking: RPC errors now return specific -320xx codes instead of generic -32603/-32600

### DIFF
--- a/crates/lib/src/error.rs
+++ b/crates/lib/src/error.rs
@@ -124,30 +124,121 @@ impl_kora_error_from!(Box<dyn StdError> => InternalServerError);
 impl_kora_error_from!(Box<dyn StdError + Send + Sync> => InternalServerError);
 impl_kora_error_from!(ProgramError => InvalidTransaction);
 
-impl From<KoraError> for RpcError {
-    fn from(err: KoraError) -> Self {
-        match err {
-            KoraError::AccountNotFound(_)
-            | KoraError::InvalidTransaction(_)
-            | KoraError::ValidationError(_)
-            | KoraError::UnsupportedFeeToken(_)
-            | KoraError::InsufficientFunds(_) => invalid_request(err),
+/// Stable numeric error codes for RPC responses following JSON-RPC 2.0 spec (-32000 to -32099 for server errors).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(into = "i32")]
+pub enum KoraErrorCode {
+    // Validation errors (-32000 to -32019)
+    InvalidTransaction = -32000,
+    ValidationError = -32001,
+    UnsupportedFeeToken = -32002,
+    InsufficientFunds = -32003,
+    InvalidRequest = -32004,
+    FeeEstimationFailed = -32005,
+    TransactionExecutionFailed = -32006,
 
-            KoraError::InternalServerError(_) | KoraError::SerializationError(_) => {
-                internal_server_error(err)
-            }
+    // Signing errors (-32020 to -32029)
+    SigningError = -32020,
 
-            _ => invalid_request(err),
-        }
+    // Auth / Rate limiting (-32030 to -32039)
+    RateLimitExceeded = -32030,
+    UsageLimitExceeded = -32031,
+    Unauthorized = -32032,
+
+    // Token / Swap (-32040 to -32049)
+    SwapError = -32040,
+    TokenOperationError = -32041,
+
+    // Account errors (-32050 to -32059)
+    AccountNotFound = -32050,
+
+    // External services (-32060 to -32069)
+    JitoError = -32060,
+    RecaptchaError = -32061,
+
+    // Internal errors (-32090 to -32099)
+    InternalServerError = -32090,
+    ConfigError = -32091,
+    SerializationError = -32092,
+    RpcError = -32093,
+}
+
+impl From<KoraErrorCode> for i32 {
+    fn from(code: KoraErrorCode) -> Self {
+        code as i32
     }
 }
 
-pub fn invalid_request(e: KoraError) -> RpcError {
-    RpcError::Call(CallError::from_std_error(e))
+impl KoraError {
+    pub fn error_code(&self) -> KoraErrorCode {
+        match self {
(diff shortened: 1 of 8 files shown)


amilz:
Thanks for this. I want to do some research in @solana/errors package before we ship this. Will report back.

raushan728:
Hi @amilz @dev-jodee any updates?

amilz:
@raushan728 still looking into something on the TS side. thanks for your patience.

amilz:
@raushan728 curious what you think about this?
https://github.com/anza-xyz/kit/pull/1576

raushan728:
> @raushan728 curious what you think about this? [anza-xyz/kit#1576](https://github.com/anza-xyz/kit/pull/1576)

Reviewed the diff as part of learning still a 2nd year student so 
some of my comments might be off. Happy to be corrected if I've misread anything.

raushan728:
Ping! Just keeping this PR active.

crates/lib/src/error.rs
```diff
@@ -338,68 +429,98 @@ mod tests {
...
-            KoraError::RpcError("test".to_string()),
-            KoraError::SigningError("test".to_string()),
-            KoraError::TransactionExecutionFailed("test".to_string()),
-            KoraError::FeeEstimationFailed("test".to_string()),
-            KoraError::SwapError("test".to_string()),
-            KoraError::TokenOperationError("test".to_string()),
-            KoraError::InvalidRequest("test".to_string()),
-            KoraError::Unauthorized("test".to_string()),
+    fn test_all_variants_have_unique_codes() {
+        use std::collections::HashSet;
```
dev-jodee: import at the top of the mod please

dev-jodee (changes requested):
small nit

crates/lib/src/error.rs
```diff
@@ -124,30 +124,121 @@ impl_kora_error_from!(Box<dyn StdError> => InternalServerError);
...
+            KoraError::InternalServerError(_) => KoraErrorCode::InternalServerError,
+            KoraError::ConfigError(_) => KoraErrorCode::ConfigError,
+            KoraError::SerializationError(_) => KoraErrorCode::SerializationError,
+            KoraError::RpcError(_) => KoraErrorCode::RpcError,
+        }
+    }
+
+    /// Returns a structured data object for the RPC error response.
+    pub fn to_json_error_data(&self) -> serde_json::Value {
+        let error_type = match self {
```
dev-jodee: Not needed to redo match, just do 

```let error_type = format!("{:?}", self.error_code());```

File: crates/lib/src/error.rs [hunk 3]
```diff
@@ -124,30 +124,121 @@ impl_kora_error_from!(Box<dyn StdError> => InternalServerError);
...
+            KoraError::RateLimitExceeded => "RateLimitExceeded",
+            KoraError::UsageLimitExceeded(_) => "UsageLimitExceeded",
+            KoraError::ConfigError(_) => "ConfigError",
+            KoraError::JitoError(_) => "JitoError",
+            KoraError::RecaptchaError(_) => "RecaptchaError",
+        };
+
+        serde_json::json!({
+            "error_type": error_type,
+            "message": self.to_string(),
```
dev-jodee: Message is already at the top level, don't think we need to duplicate it ?

sdks/ts/src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+    SerializationError = -32092,
+    RpcError = -32093,
+}
+
+/**
+ * Structured data returned in the JSON-RPC error object's `data` field.
+ */
+export interface KoraErrorData {
+    error_type: string;
+    message: string;
```
dev-jodee: think this is duplicate, its same as Error.message?

sdks/ts/src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Represents a structured error returned by the Kora paymaster service.
+ */
+export class KoraError extends Error {
+    /** The stable numeric error code */
+    public readonly code: number;
```
dev-jodee: should be KoraErrorCode ?

crates/lib/src/error.rs
```diff
@@ -124,30 +124,121 @@ impl_kora_error_from!(Box<dyn StdError> => InternalServerError);
...
-            | KoraError::ValidationError(_)
-            | KoraError::UnsupportedFeeToken(_)
-            | KoraError::InsufficientFunds(_) => invalid_request(err),
-
-            KoraError::InternalServerError(_) | KoraError::SerializationError(_) => {
-                internal_server_error(err)
-            }
+/// Stable numeric error codes for RPC responses following JSON-RPC 2.0 spec (-32000 to -32099 for server errors).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+#[serde(into = "i32")]
```
dev-jodee: dont think this is needed ? you already do as i32

dev-jodee (changes requested):

crates/lib/src/error.rs
raushan728: Done, also removed the `From<KoraErrorCode> for i32` impl since it was only there for this serde attribute nothing else was using it.

crates/lib/src/error.rs
raushan728: Done

dev-jodee (approved):

dev-jodee:
Will merge this one because it's old, but for future work @raushan728 we'll enforce signed commits :) thanks again for the amazing work

raushan728:
Thanks! I actually didn't know about signed commits until now, so I'll get that set up.

Just to be clear, I have a few other open PRs here that already contain multiple unsigned commits. If I make follow up changes after review, is it enough for the new commits to be signed, or would you prefer that I go back and re-sign the existing commits in those PRs as well?

dev-jodee:
It's fine for those existing PRs, but for future work it'll be a requirement :)

PR #450, opened 2026-04-23
chore(ts-sdk): bump kit to 6.8 and plugins to 0.10

amilz:
### Kit + plugins bump
- Bumps `@solana/kit` `^6.3.0` → `^6.8.0` and kit plugins `^0.7.0` → `^0.10.0` (peer deps).
- Replaces deprecated APIs: `createEmptyClient` → `createClient`, `rpc(url)` → `solanaRpcConnection(url)`, `@solana/kit-plugin-payer` → `@solana/kit-plugin-signer`.
- Migrates test setup off deprecated `@solana/kit-client-litesvm` to composed `createClient() + payer() + litesvm()` using `@solana/kit-plugin-litesvm`.
- Bumps `@solana-program/token` `^0.12.0` → `^0.13.0` and SDK version to `0.3.0-beta.1`.

### New `kora()` bundle plugin
- Adds `kora()` — a single plugin that mirrors `solanaRpc`'s shape and composes RPC + payer + Kora RPC + transaction planning/execution. User signer is read from `client.identity` (install via `@solana/kit-plugin-signer`); Kora's fee-payer address is fetched from the node and installed internally as a noop signer.
- Recommended usage:
  ```ts
  import { createClient } from '@solana/kit';
  import { identity } from '@solana/kit-plugin-signer';
  import { kora } from '@solana/kora';

  const client = await createClient()
    .use(identity(userSigner))
    .use(await kora({ endpoint, rpcUrl, feeToken }));
  ```
- Adds `KoraBundleConfig` type (same as `KoraKitClientConfig` minus `feePayerWallet`).
- Refactors `createKoraTransactionPlanExecutor` to take `userSigner` as an explicit arg.

### Deprecations
- `createKitKoraClient` and `KoraKitClientConfig` marked `@deprecated`; wrapper now delegates to `identity()` + `kora()`. Scheduled for removal in `0.4.0`.

## Test Plan
- [x] `pnpm run build` (type-check clean)
- [x] `pnpm run lint`
- [x] `pnpm run test:ci:unit` — 39/39 passed
- [x] `npx jest test/kit-client.test.ts` — 20/20 passed (incl. new `kora() bundle plugin` describe)
- [x] `npx jest test/plugin.test.ts` — 23/23 passed
- [x] `just integration-test --filter typescript_basic` — passed (~33s)

## Breaking Changes
- Consumers of the kit plugin pattern must update: `createEmptyClient()` → `createClient()`, and swap `@solana/kit-plugin-payer` for `@solana/kit-plugin-signer`.
- `createKitKoraClient` still works but emits a deprecation hint; migrate to `identity()` + `kora()` before `0.4.0`.

Closes DEV-420


amilz:
fyi @lorisleiva

sdks/ts/src/kit/index.ts
lorisleiva: If you move that logic (and the lines after) inside the plugin function, then you remove the need for consumers to call `await` on the API.

```diff
  const client = await createClient()
    .use(identity(userSigner))
-   .use(await kora({ endpoint, rpcUrl, feeToken }));
+   .use(kora({ endpoint, rpcUrl, feeToken }));
```

The logic will still only be called once upon client initialisation.

sdks/ts/src/kit/index.ts
```diff
@@ -68,46 +67,74 @@ export async function createKitKoraClient(config: KoraKitClientConfig) {
...
+            koraPaymentAddress(paymentAddr),
+            transactionPlannerPlugin(koraTransactionPlanner),
+            transactionPlanExecutorPlugin(koraTransactionPlanExecutor),
+            planAndSendTransactions(),
+        );
+    };
+}
+
+/** The type returned by {@link createKitKoraClient}. */
+export type KoraKitClient = Awaited<ReturnType<typeof createKitKoraClient>>;
```
lorisleiva: I guess this can also be deprecated now?

sdks/ts/test/integration.test.ts
```diff
@@ -425,16 +427,23 @@ describe(`KoraClient Integration Tests (${AUTH_ENABLED ? 'with auth' : 'without
...
+            async function buildFreeClient() {
                 const koraRpcUrl = process.env.KORA_RPC_URL || 'http://127.0.0.1:8080';
-                freeClient = await createKitKoraClient({
-                    endpoint: koraRpcUrl,
-                    rpcUrl: process.env.SOLANA_RPC_URL || 'http://127.0.0.1:8899',
-                    feeToken: usdcMint,
-                    feePayerWallet: testWallet,
-                });
+                return createClient()
+                    .use(identity(testWallet))
```
lorisleiva, line 440: It might be worth adding a test that uses `signer(testWallet)` instead of `identity(testWallet)` because the former sets both `client.identity` and `client.payer` and it would be nice to ensure `client.payer` gets overridden by the Kora payer as expected.

In general, we're gonna be leaning into the "signer" variants (e.g. `generatedSigner`, `signerFromFile`, etc.) in our documentation because it sets both to the same signer which is what you'd want in most cases.

sdks/ts/test/setup.ts
```diff
@@ -126,7 +128,11 @@ async function setupTestSuite(): Promise<TestSuite> {
             : undefined;
 
     const { testWallet, usdcMint, destinationAddress } = await createKeyPairSigners();
-    const client = await createClient({ payer: testWallet }).use(tokenProgram()).use(associatedTokenProgram());
+    const client = await createClient()
+        .use(payer(testWallet))
+        .use(litesvm())
+        .use(tokenProgram())
+        .use(associatedTokenProgram());
```
lorisleiva, line 135: I know this is a bit more verbose but I'm actually really glad we landed on that design. There's a lot less hidden away from the consumer and it becomes clearer what they need to do to swap a payer or an RPC/SVM provider. 👌

lorisleiva (approved):
Nice! I've left a few minor comments but looks good to me, thank you!

dev-jodee (dismissed):

sdks/ts/test/integration.test.ts
amilz: test currently skipped. dep on:
- anza-xyz/kit#1569
- anza-xyz/kit-plugins#204

dev-jodee (dismissed):

dev-jodee (approved):

PR #456, opened 2026-04-28
fix(validator): enforce fee-payer policy for multisig co-signers

raushan728:
The fee-payer policy validator only checked direct owner equality, so a transaction could use the fee payer as a multisig co-signer on Burn, Approve, Transfer, CloseAccount, Revoke, SetAuthority, MintTo, FreezeAccount, ThawAccount, and Reallocate instructions without triggering the policy check.

**Changes:**
- Added `multisig_signers` field to 8 `ParsedSPLInstructionData` variants
- Updated SPL and Token-2022 parsers to extract co-signers via `extract_multisig_signers`
- Extended `validate_spl!` macro to check both direct authority and multisig signers
- Extended `SplTokenReallocate` to extract and check trailing multisig co-signers
- Fixed SetAuthority and ApproveChecked skip offsets for correct co-signer extraction
- Added 5 integration tests covering Burn, Approve, Transfer, SetAuthority, and Reallocate bypass scenarios

**Tests:** 24/24 passing

---

<a href="https://app.devin.ai/review/solana-foundation/kora/pull/456" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/lib/src/transaction/instruction_util.rs
+++ b/crates/lib/src/transaction/instruction_util.rs
@@ -205,6 +205,7 @@ pub enum ParsedSPLInstructionData {
         account: Pubkey,
         payer: Pubkey,
         owner: Pubkey,
+        multisig_signers: Vec<Pubkey>,
         is_2022: bool,
     },
     SplTokenInitializePausable {
@@ -3206,6 +3207,7 @@ impl IxUtils {
                                     account: instruction.accounts[instruction_indexes::spl_token_reallocate::ACCOUNT_INDEX].pubkey,
                                     payer: instruction.accounts[instruction_indexes::spl_token_reallocate::PAYER_INDEX].pubkey,
                                     owner: instruction.accounts[instruction_indexes::spl_token_reallocate::OWNER_INDEX].pubkey,
+                                    multisig_signers: Self::extract_multisig_signers(instruction, 4),
                                     is_2022: true,
                                 });
                         }
--- a/crates/lib/src/transaction/instruction_util.rs
+++ b/crates/lib/src/transaction/instruction_util.rs
@@ -138,32 +138,38 @@ pub enum ParsedSPLInstructionData {
     // Includes burn and burn with seed
     SplTokenBurn {
         owner: Pubkey,
+        multisig_signers: Vec<Pubkey>,
         is_2022: bool,
     },
     // Includes close account
     SplTokenCloseAccount {
         owner: Pubkey,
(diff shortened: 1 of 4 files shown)


raushan728:
@dev-jodee could you trigger a re-review from the bots? I've pushed fixes and I am not able to request review form bots.

tests/fee_payer_policy/fee_payer_policy_violations.rs
```diff
@@ -773,3 +773,467 @@ async fn test_thaw_account_token2022_policy_violation() {
...
+        )
+        .with_instruction(burn_ix)
+        .build()
+        .await
+        .expect("Failed to build tx");
+
+    let result =
+        ctx.rpc_call::<serde_json::Value, _>("signTransaction", rpc_params![bypass_tx]).await;
+
+    assert!(result.is_err(), "fee payer as multisig co-signer should be rejected");
```
dev-jodee: Can we assert the actual error message here instead of just `is_err()`?

This could pass for a payment/setup/signature failure, without proving the fee-payer policy rejected the multisig signer case. The other tests in this file use `error.assert_contains_message(...)`; same pattern would work here.

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -184,11 +190,13 @@ pub enum ParsedSPLInstructionData {
...
+        multisig_signers: Vec<Pubkey>,
         is_2022: bool,
     },
     // ThawAccount
     SplTokenThawAccount {
         freeze_authority: Pubkey,
+        multisig_signers: Vec<Pubkey>,
         is_2022: bool,
     },
     // Token2022 Reallocate
```
dev-jodee, line 203: `Token2022 Reallocate` still has this bypass.

The account layout is `[account, payer, system_program, owner, ...signers]`. We only store/check `payer` and `owner` here, so if `owner` is a multisig and the fee payer is one of the trailing signer accounts, the policy check won’t catch it.

Can we add the trailing signers to `SplTokenReallocate` and reject when they contain the fee payer too?

raushan728: Fixed - added multisig_signers to SplTokenReallocate, extract trailing signers from index 4, and added a regression test for this case.

dev-jodee:
thanks for the work as always @raushan728 !

PR #463, opened 2026-05-01
fix(usage_limit): atomic usage limit enforcement across concurrent requests

raushan728:
The usage limiter was doing a read-then-increment in two separate steps, so concurrent requests could both pass the limit check before either one incremented.

Added `check_and_increment` to the store trait  single mutex lock for in-memory, Lua script for Redis. Pre-checks all rules before incrementing; denied requests may still consume quota from earlier rules under concurrent load. TTL only set on first increment within a window.

### DIFF
--- a/crates/lib/src/usage_limit/usage_store.rs
+++ b/crates/lib/src/usage_limit/usage_store.rs
@@ -2,6 +2,7 @@ use std::{collections::HashMap, sync::Mutex};
 
 use async_trait::async_trait;
 use deadpool_redis::{Connection, Pool};
+use once_cell::sync::Lazy;
 use redis::AsyncCommands;
 
 use crate::{error::KoraError, sanitize_error};
@@ -18,9 +19,35 @@ pub trait UsageStore: Send + Sync {
     /// Get current usage count for a key (returns 0 if not found)
     async fn get(&self, key: &str) -> Result<u32, KoraError>;
 
+    /// Atomic check and increment: check if (current + delta) <= max, and increment if so.
+    /// Returns true if allowed and incremented, false if denied.
+    async fn check_and_increment(
+        &self,
+        key: &str,
+        delta: u64,
+        max: u64,
+        expiry: Option<u64>,
+    ) -> Result<bool, KoraError>;
+
     /// Clear all usage data (mainly for testing)
     async fn clear(&self) -> Result<(), KoraError>;
 }
+
+/// Atomically checks the limit and increments - sets TTL only on first increment.
+static CHECK_AND_INCREMENT_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
+    redis::Script::new(
+        r"
+            local current = redis.call('GET', KEYS[1])
+            local count = current and tonumber(current) or 0
+            if count + tonumber(ARGV[1]) > tonumber(ARGV[2]) then return 0 end
+            redis.call('INCRBY', KEYS[1], ARGV[1])
+            if ARGV[3] ~= '0' and redis.call('TTL', KEYS[1]) < 0 then
+                redis.call('EXPIREAT', KEYS[1], ARGV[3])
+            end
+            return 1
+            ",
+    )
+});
 
 /// Redis-based implementation for production
 pub struct RedisUsageStore {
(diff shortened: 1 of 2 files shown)


crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -289,51 +282,16 @@ impl UsageTracker {
...
-                // Subsequent increments without resetting expiry
-                for _ in 1..increment_count {
-                    self.store.increment(&key).await?;
-                }
-            } else {
-                for _ in 0..increment_count {
-                    self.store.increment(&key).await?;
-                }
-            }
+            log::debug!("[rule] User {} {}: (atomic check+inc)", ctx.user_id, rule.description());
```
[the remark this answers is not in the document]
raushan728: Known limitation noted in the doc comment and test. Full fix needs rollback support, tracking as a follow-up.

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -289,51 +282,16 @@ impl UsageTracker {
...
-                    reason: format!(
-                        "User {} exceeded {} limit: {}/{}",
-                        ctx.user_id,
-                        rule.description(),
-                        new_count,
-                        rule.max()
-                    ),
+                    reason: format!("User {} exceeded {} limit", ctx.user_id, rule.description()),
                 });
             }
```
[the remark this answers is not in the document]
raushan728: Same acknowledged in doc comment and test.

raushan728:
Just keeping this PR active

crates/lib/src/usage_limit/usage_store.rs
```diff
@@ -90,6 +100,45 @@ impl UsageStore for RedisUsageStore {
...
+    async fn check_and_increment(
+        &self,
+        key: &str,
+        delta: u64,
+        max: u64,
+        expiry: Option<u64>,
+    ) -> Result<bool, KoraError> {
+        let mut conn = self.get_connection().await?;
+
+        let script = redis::Script::new(
```
dev-jodee: I think this script should only be created once, and would benefit from a comment explaining what it does

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -1028,4 +1063,92 @@ mod tests {
...
+                allowed_count += 1;
+            }
+        }
+
+        assert_eq!(allowed_count, 2);
+
+        let lifetime_key = "kora:tx:multi-rule-concurrent-user";
+        let lifetime_count = store.get(lifetime_key).await.unwrap();
+        // denied requests may have consumed quota from earlier rules
+        assert!(lifetime_count >= allowed_count as u32);
```
dev-jodee, line 1152: should also assert an upper bound?

dev-jodee (changes requested):

crates/lib/src/usage_limit/usage_store.rs
raushan728: Will move this to `once_cell::sync::Lazy` so the script is initialized once and reused across calls.

crates/lib/src/usage_limit/usage_tracker.rs
raushan728: Will add assert!(lifetime_count <= 10) as upper bound.

crates/lib/src/usage_limit/usage_store.rs
raushan728: Done

crates/lib/src/usage_limit/usage_tracker.rs
raushan728: Done

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -265,75 +266,61 @@ impl UsageTracker {
...
                 continue;
             }
 
             let key = rule.storage_key(&ctx.user_id, ctx.timestamp);
+            let expiry =
+                rule.window_seconds().filter(|&w| w > 0).map(|w| (ctx.timestamp / w + 1) * w);
+            let max = rule.max();
+            let description = rule.description();
 
             let current = self.store.get(&key).await?;
```
dev-jodee, line 300: Can we add a comment here saying it's a fath path to avoid hitting check and icnrement for keys that are already over the limit, jut so it's clear (keep it concise)

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -265,75 +266,61 @@ impl UsageTracker {
...
-                    self.store.increment(&key).await?;
-                }
+        for (key, increment_count, max, expiry, description) in pending_increments {
+            if !self.store.check_and_increment(&key, increment_count, max, expiry).await? {
+                return Ok(LimiterResult::Denied {
+                    reason: format!("User {} exceeded {} limit", ctx.user_id, description),
+                });
             }
+
+            log::debug!("[rule] User {} {}: (atomic check+inc)", ctx.user_id, description);
```
dev-jodee: we can remove that

crates/lib/src/usage_limit/usage_store.rs
```diff
@@ -18,10 +19,36 @@ pub trait UsageStore: Send + Sync {
...
 
+/// Atomically checks the limit and increments - sets TTL only on first increment.
+static CHECK_AND_INCREMENT_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
+    redis::Script::new(
+        r"
+            local current = redis.call('GET', KEYS[1])
+            local count = current and tonumber(current) or 0
+            if count + tonumber(ARGV[1]) > tonumber(ARGV[2]) then return 0 end
+            redis.call('INCRBY', KEYS[1], ARGV[1])
+            if ARGV[3] ~= '0' and redis.call('TTL', KEYS[1]) < 0 then
```
dev-jodee, line 45: Can we add to the comment that the 0 is the sentinel value for lua

dev-jodee (changes requested):
small nits

dev-jodee (approved):

PR #464, opened 2026-05-03
feat(validator): cross-cluster mint validation

raushan728:
Warns the operator if a configured mint is not found on the connected cluster.

Probes configured endpoints concurrently (5s timeout). Possible outcomes:
- mint found on another cluster reports which one, flags possible cluster mismatch
- mint confirmed absent on all probed clusters lists checked clusters
- all probes failed or timed out warns that check was inconclusive
- connected cluster unreachable warns that check was skipped

Opt-in via `cross_cluster_check = true` in kora.toml disabled by default to avoid sending mint addresses to public RPC endpoints. Probe endpoints are configurable with public Solana clusters as defaults.

### DIFF
--- a/crates/lib/src/validator/config_validator.rs
+++ b/crates/lib/src/validator/config_validator.rs
@@ -32,6 +32,14 @@ use spl_token_interface::ID as SPL_TOKEN_PROGRAM_ID;
 
 const MIN_SIGN_TIMEOUT_SECONDS: u64 = 1;
 const HIGH_SIGN_MAX_RETRIES_WARNING_THRESHOLD: u32 = 10;
+const CROSS_CLUSTER_TIMEOUT_SECS: u64 = 5;
+
+enum ProbeOutcome {
+    Found(Vec<String>),
+    NotFound,
+    /// RPC error or timeout cannot conclude the mint is absent on this cluster.
+    Failed,
+}
 
 pub struct ConfigValidator {}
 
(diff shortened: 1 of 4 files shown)


raushan728:
@greptile re-review

amilz:
Thanks for this. Couple of thoughts. WDYT?
1. Wonder if we should have a `[validation] cross_cluster_check = false` config flag (default off) so teams that want their mint list private don't send their mints to public RPC endpoint. 
2. Catch-all `_ => (*cluster_name, vec![])` collapses every failure mode. If many mints, public RPC may rate limit and give false errors. Wonder if we should have something like `enum ProbeOutcome { Found, NotFound, Failed }`
3. Maybe need a timeout to prevent an infinite hang?
4. Should we mock results for `cross_cluster_validation_test.rs` to ensure we don't create CI issues w/ our tests hitting these clusters
5. Do we need a `found on: devnet message` test case?

raushan728:
Thanks for the review!
1. will add cross_cluster_check = false config flag, default off for privacy
2. will introduce ProbeOutcome to distinguish rate limit failures from actual missing mints
3. Timeout is already in place — CROSS_CLUSTER_TIMEOUT_SECS = 5 per cluster probe
4. will mock the RPC responses in tests to avoid CI flakiness
5. Will add a test case for the found on: devnet warning path

raushan728:
@amilz all done!

kora.toml
```diff
@@ -53,6 +53,7 @@ max_allowed_lamports = 1000000
 max_signatures = 10
 price_source = "Mock"
 allow_durable_transactions = false
+# cross_cluster_check = false  # opt-in: probes mainnet/devnet/testnet to detect cluster mismatches
```
dev-jodee: can just leave it uncommented with false as default, does the same "by default not enabled"

crates/cli/src/main.rs
```diff
@@ -167,6 +167,32 @@ async fn main() -> Result<(), KoraError> {
                         }
                     }
 
+                    let mut startup_warnings: Vec<String> = Vec::new();
```
dev-jodee: I think this should go in the `validate_with_result_and_signers` as well, is there a reason you put it at this level instead of in the function?

crates/lib/src/validator/config_validator.rs
```diff
@@ -32,6 +32,24 @@ use spl_token_interface::ID as SPL_TOKEN_PROGRAM_ID;
 
 const MIN_SIGN_TIMEOUT_SECONDS: u64 = 1;
 const HIGH_SIGN_MAX_RETRIES_WARNING_THRESHOLD: u32 = 10;
+const CROSS_CLUSTER_TIMEOUT_SECS: u64 = 5;
+
+const CLUSTER_RPC_URLS: &[(&str, &str, &str)] = &[
+    (
+        "mainnet-beta",
+        "https://api.mainnet-beta.solana.com",
```
dev-jodee: those should be a configurable endpoint, for all envs, as there's a high chance of being rate limited on the public ones

dev-jodee (commented):
couple of questions / comments

kora.toml
raushan728: Makes sense, will uncomment it with false as default

crates/cli/src/main.rs
raushan728: The reason it's at startup level is that `validate_with_result_and_signers` is called with `skip_rpc_validation: true` at startup  so moving it inside the function would require either flipping that flag (which enables all RPC validation, not just this check) or adding a separate parameter just for cross-cluster. We intentionally kept it isolated to avoid side effects. To move it if you think a dedicated parameter is the right approach what's your preference?

crates/lib/src/validator/config_validator.rs
raushan728: Rate limiting is valid. We could let operators override the probe endpoints in `kora.toml` but adds config surface for an already opt-in check. Would keeping defaults with an optional override work, or do you want them always configurable?

dev-jodee: Yeah optional with defaults works

crates/cli/src/main.rs
dev-jodee: Can we make it always check even if skip rpc validate is true ? in the validate_with_result_and_signers call ?

raushan728: Sure, will add a dedicated parameter to validate_with_result_and_signers so cross-cluster check always runs regardless of skip_rpc_validation.

crates/lib/src/validator/config_validator.rs
raushan728: Will add optional override in kora.toml with public endpoints as defaults

raushan728: Done

crates/cli/src/main.rs
raushan728: Done

dev-jodee (approved):

PR #465, opened 2026-05-04
Add "All" wildcard sentinel for allowed_programs

isardmart:
## Summary                       
                                                                            
Adds an explicit wildcard sentinel `"*"` to `allowed_programs` so operators can opt out of program-level allowlisting when policy is enforced upstream.                                                                 
                                                      
The default behavior is unchanged. Operators only enter wildcard mode by explicitly setting `allowed_programs = ["*"]` in their config. 
                                                                            
 ## Motivation                                                  
                                                                            
Hosted paymaster operators (e.g. Openfort) authorize which programs each customer can sponsor in their policy engine. Kora's static `allowed_programs` forces a duplicated, manually maintained allowlist that must be edited in `kora.toml` and redeployed every time a customer adds a program. Drift between the two layers surfaces to end users as `-32000` errors that look like infrastructure failures.                            
                                                                            
This lets operators delegate the program-level decision upstream while still relying on `fee_payer_policy`, `disallowed_accounts`, `max_signatures`, and `max_allowed_lamports` to bound drainage risk.             
                                                                            
  ## Design                                                      
                                                                            
  `"*"` was chosen over an empty-list semantic so existing deployments with `allowed_programs = []` keep their current "this will block all transactions" warning, no silent migration.               
                                                                            
  ## Changes                                                                
                                                                            
  - `crates/lib/src/validator/transaction_validator.rs`                     
    - New `allow_all_programs: bool` field, set in `TransactionValidator::new` when any entry in `allowed_programs` equals `"*"`.         
    - Sentinel filtered out before parsing remaining entries as `Pubkey`s, so mixed lists like `["*", "Token..."]` don't fail at parse time.     
    - `validate_programs` short-circuits with `Ok(())` when wildcard is set.
    - Same short-circuit applied to the `SystemCreateAccount` owner check  inside `validate_fee_payer_usage`.                                    
  - `crates/lib/src/validator/config_validator.rs`                          
    - New `WILDCARD MODE` warning when `"*"` is present.                    
    - Original empty-list warning preserved for the deny-all case.          
  - `kora.toml`                                                             
    - Sample config switched to `["*"]` to exercise the new path.


kora.toml
```diff
@@ -54,14 +54,7 @@ max_signatures = 10
...
 
-allowed_programs = [
-    "11111111111111111111111111111111",              # System Program
-    "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",   # Token Program
-    "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",  # Associated Token Program
-    "AddressLookupTab1e1111111111111111111111111",   # Address Lookup Table Program
-    "ComputeBudget111111111111111111111111111111",   # Compute Budget Program
-    # "L2TExMFKdjpN9kozasaurPirfHy9P8sbXoAN1qA3S95", # Lighthouse Program (uncomment when lighthouse enabled)
-]
+allowed_programs = ["*"]
```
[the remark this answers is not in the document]
dev-jodee: Yeah I asgree with this, default shouldn't be wildcard, we try to give strict defaults just so people don't operate a node with wide permissions by mistake

crates/lib/src/validator/transaction_validator.rs
```diff
@@ -38,10 +39,13 @@ impl TransactionValidator {
     pub fn new(config: &Config, fee_payer_pubkey: Pubkey) -> Result<Self, KoraError> {
         let config = &config.validation;
 
-        // Convert string program IDs to Pubkeys
+        // "*" is a wildcard sentinel, filtered out before parsing remaining entries as Pubkeys.
+        let allow_all_programs = config.allowed_programs.iter().any(|addr| addr == "*");
```
dev-jodee: I think we should make it that if there's only 1 and it's the wildcard, then its considered wildcard. If there's more than one (even if one is the wildcard), it shouldn't accept all, might be a mistake from the operator.

isardmart: makes sense, changing those 2 things right now

isardmart:
Components touched (across both commits):     
                                                
  - kora.toml                                   
    - Removed the wildcard from the default `allowed_programs`; ships the explicit list of standard programs again. 
  - crates/lib/src/config.rs                    
    - Added `ALLOWED_PROGRAMS_WILDCARD = "*"` const.                                        
    - Added `AllowedProgramsMode` enum: `Wildcard` | `StrayWildcard` | `Explicit`.    
    - Added `ValidationConfig::allowed_programs_mode()` , single classifier consumed by both validators.
  - crates/lib/src/validator/transaction_validator.rs                                         
    - `TransactionValidator::new` now dispatches on `allowed_programs_mode()`:         
      `Wildcard` → `allow_all_programs = true`;
     `StrayWildcard` → hard error;                 
      `Explicit` → parse list as before.
  - crates/lib/src/validator/config_validator.rs
    - Same match in `validate_with_warnings`:
  `StrayWildcard` pushes an error, `Wildcard` keeps the warning.
    - In wildcard mode, skips the empty/missing-System/missing-Token warnings, the lighthouse "must be in         
  allowed_programs" error, and `warn_unvalidated_programs`.            
                                         
  Architectural change:                         
  Wildcard detection moved from a scattered `any(== "*")` check to a single 3-state classifier on `ValidationConfig`. The "* mixed with other entries" case is now an explicit, rejected state instead of silently enabling wildcard mode.

amilz:
Looks good, but inclined to mirror the `SplTokenConfig::All` pattern from #175 instead of a `"*"` sentinel:

```toml
allowed_programs = "All"
# or
allowed_programs = ["Sys111…", "Token…"]
```
@dev-jodee wdyt?

- Makes `StrayWildcard` unrepresentable — drops the runtime check, the second error path in `TransactionValidator::new`, and one of the new tests.
- Consistency with `allowed_spl_paid_tokens`; operators learn one pattern.
- I think the `*` may regress some of the existing base58 config validation searches

isardmart:
## Replace `"*"` wildcard with `ProgramsConfig::All` enum 
                                                            
 Mirrors `SplTokenConfig::All` pattern per review. Drops`StrayWildcard` runtime check + base58 special-case.      
                                                            
  ```rust                                                   
  pub enum ProgramsConfig {                                 
      All,                                                  
      #[serde(untagged)]                                    
      Allowlist(Vec<String>),                               
  }                                      
```
  Bundle validator bug fix: setting allowed_programs = "*" previously caused the bundle validator to reject every program (empty HashSet from parse_pubkey_set). Now threads allow_all_programs: bool, mirroring transaction_validator.rs. disallowed_accounts still enforced under wildcard.

  TOML:
  allowed_programs = "All"            # wildcard
  allowed_programs = ["...", "..."]   # explicit (unchanged)
                                                            
  Tests: 784/784 pass. 
New: bundle wildcard accepts arbitrary; bundle wildcard still rejects disallowed.

File: crates/lib/src/config.rs [hunk 3]
```diff
@@ -117,11 +117,50 @@ impl SplTokenConfig {
...
+        }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct ValidationConfig {
     pub max_allowed_lamports: u64,
     pub max_signatures: u64,
-    pub allowed_programs: Vec<String>,
+    pub allowed_programs: ProgramsConfig,
```
dev-jodee, line 190: `allowed_programs` is no longer always an array. With `ProgramsConfig::All`, the Rust API can serialize this as `"All"`, but the OpenAPI spec and TS SDK still model it as `string[]`, and the Kit wrapper calls `.map(...)` unconditionally. A wildcard-configured server can make `getConfig()` throw at runtime. Please update the OpenAPI/TS contract and handle both `"All"` and `string[]`, or normalize the RPC response so it remains array-compatible.

crates/lib/src/config.rs
```diff
@@ -117,11 +117,50 @@ impl SplTokenConfig {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
```
dev-jodee: This adds a new serialized config shape, but there is no test that loads `allowed_programs = "All"` from TOML or checks the JSON returned by `getConfig`. The validator tests cover direct `ProgramsConfig::All` assignment, but not the user-facing config/API path where this can break clients. Please add coverage for TOML parsing and `getConfig` serialization.

dev-jodee:
The OpenAPI specs and TS types still document `allowed_programs` as `string[]`, but the Rust type can now serialize as `"All"`. These generated/static contract files need to be updated with the new union shape, and the SDK tests should include the wildcard response case.

dev-jodee (changes requested):
couple of small things to update

dev-jodee (dismissed):

dev-jodee (dismissed):

dev-jodee (approved):

PR #469, opened 2026-05-07
fix(validator): skip empty key vars

abelmarnk:
The changes in this PR would cause empty keys to be treated as not present, and would follow the path of issuing warnings when those keys are absent.

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -696,6 +696,17 @@ impl Default for AuthConfig {
     }
 }
 
+impl AuthConfig {
+    pub(crate) fn normalize_optional_secret(value: Option<String>) -> Option<String> {
+        value.filter(|value| !value.is_empty())
+    }
+
+    pub(crate) fn has_auth(&self) -> bool {
+        self.api_key.as_deref().is_some_and(|key| !key.is_empty())
+            || self.hmac_secret.as_deref().is_some_and(|key| !key.is_empty())
+    }
+}
+
 impl Config {
     pub fn load_config<P: AsRef<Path>>(path: P) -> Result<Config, KoraError> {
         let contents = fs::read_to_string(path).map_err(|e| {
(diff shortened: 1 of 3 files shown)


crates/lib/src/config.rs
```diff
@@ -696,6 +696,17 @@ impl Default for AuthConfig {
     }
 }
 
+impl AuthConfig {
+    pub fn normalize_optional_secret(value: Option<String>) -> Option<String> {
```
raushan728: nit: pub(crate) enough here since it's only used within the crate?

crates/lib/src/rpc_server/server.rs
```diff
@@ -34,7 +35,8 @@ pub struct ServerHandles {
 
 // We'll always prioritize the environment variable over the config value
 fn get_value_by_priority(env_var: &str, config_value: Option<String>) -> Option<String> {
-    std::env::var(env_var).ok().or(config_value)
+    AuthConfig::normalize_optional_secret(std::env::var(env_var).ok())
```
raushan728, line 38: nit: if more sources get added later, calling `normalize_optional_secret` on each manually could get repetitive

abelmarnk: what would you suggest?

raushan728: maybe normalize at the end after all sources are resolved?

```rust
let raw = std::env::var(env_var).ok().or(config_value);
AuthConfig::normalize_optional_secret(raw)

abelmarnk: thanks for the suggestion, i think that may conflict with what the function does, in the case where `env_var` was empty, it would return `None` and ignore the config value.

raushan728: ah right, makes sense! thanks for clarifying

crates/lib/src/config.rs
abelmarnk: resolved [in](https://github.com/solana-foundation/kora/pull/469/changes/fa183b3c754b44100bbdaabcd57cffe24ad27dba#diff-f75f5a0df893f7855b23ce262dc751ad4d339d34cbd44c3cd187a432382bb066R700)

dev-jodee (approved):

PR #471, opened 2026-05-07
Caching for exchange rates

n00m4d:
# What
This PR adds Redis caching for token prices used during fee calculation.

# Why
Price oracle API calls (e.g. Jupiter) can be expensive and add latency. For back-to-back requests like `signTransaction` or `signAndSendTransaction` landing within seconds of each other, the price doesn't change meaningfully — refetching it every time is wasteful.

# How
Token prices are stored in Redis with a configurable TTL (`price_ttl`, default 30s). On a cache miss, only the missing mints are fetched from the oracle and written back. A per-mint singleflight lock prevents concurrent requests from triggering duplicate oracle calls for the same token. The price oracle is initialized once as a process-wide singleton so its HTTP connection pool is reused. Setting `price_ttl = 0` disables price caching independently of the global cache toggle, so account caching can remain on.

### DIFF
--- a/crates/lib/src/cache.rs
+++ b/crates/lib/src/cache.rs
@@ -10,7 +10,7 @@ use tokio::sync::{Mutex, OnceCell};
 use crate::{
     config::Config,
     error::KoraError,
-    oracle::{get_price_oracle, RetryingPriceOracle, TokenPrice},
+    oracle::{get_price_oracle, PriceSource, RetryingPriceOracle, TokenPrice},
     sanitize_error,
 };
 
@@ -37,7 +37,7 @@ static CACHE_POOL: OnceCell<Option<Pool>> = OnceCell::const_new();
 /// Process-wide price oracle. Held as an `Arc` so the inner `reqwest::Client`
 /// (and its connection pool) is reused across all cache misses instead of
 /// being rebuilt per request.
-static PRICE_ORACLE: OnceCell<Arc<RetryingPriceOracle>> = OnceCell::const_new();
+static PRICE_ORACLE: OnceCell<(PriceSource, Arc<RetryingPriceOracle>)> = OnceCell::const_new();
 
 /// Per-mint locks used to coalesce concurrent oracle fetches for the same
 /// price key (singleflight). Without this, a TTL expiry under load triggers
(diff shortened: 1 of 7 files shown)


crates/lib/src/cache.rs
```diff
@@ -343,6 +367,220 @@ impl CacheUtil {
...
+        }
+
+        // If cache is disabled globally, pool not initialized, or price caching
+        // is opted out via `price_ttl = 0`, go straight to the oracle.
+        if !Self::is_cache_enabled(config) || config.kora.cache.price_ttl == 0 {
+            return Self::get_price_oracle_singleton(config)
+                .await?
+                .get_token_prices(mint_addresses)
+                .await;
+        }
```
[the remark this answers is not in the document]
n00m4d: That matches the pre-PR behavior (before this change every price fetch hit the oracle directly with no coalescing), so this PR isn't introducing a regression.
But it might be fixed/implemented in other PR as a separate fix though.

kora.toml
```diff
@@ -34,6 +34,7 @@ enabled = false                    # Enable/disable caching (set to true with ur
 url = "redis://localhost:6379"    # Redis connection URL (uncomment and set when enabling cache)
 default_ttl = 300                  # Default TTL in seconds (5 minutes)
 account_ttl = 60                   # Account data TTL in seconds (1 minute)
+price_ttl = 30                     # Token price TTL in seconds (set to 0 to disable price caching while keeping account caching on)
```
dev-jodee: caching should be OFF by default, risky to cache price in the crypto world

crates/lib/src/constant.rs
```diff
@@ -49,6 +49,7 @@ pub const DEFAULT_METRICS_SCRAPE_INTERVAL: u64 = 60;
 // Cache
 pub const DEFAULT_CACHE_DEFAULT_TTL: u64 = 300; // 5 minutes
 pub const DEFAULT_CACHE_ACCOUNT_TTL: u64 = 60; // 1 minute for account data
+pub const DEFAULT_CACHE_PRICE_TTL: u64 = 30; // 30 seconds for token price data
```
dev-jodee: 0 as well here

crates/lib/src/config.rs
```diff
@@ -564,6 +564,13 @@ pub struct CacheConfig {
     pub default_ttl: u64,
     /// TTL for account data cache in seconds
     pub account_ttl: u64,
+    /// TTL for token price data cache in seconds
+    #[serde(default = "default_price_ttl")]
+    pub price_ttl: u64,
+}
+
+fn default_price_ttl() -> u64 {
```
dev-jodee: why do we need a specific function just for default? we can use ```#[serde(default)]``` directly

crates/lib/src/cache.rs
```diff
@@ -343,6 +367,239 @@ impl CacheUtil {
...
+            Err(e) => {
+                log::warn!("Failed to read prices from cache, falling back to oracle: {e}");
+                return Self::get_price_oracle_singleton(config)
+                    .await?
+                    .get_token_prices(mint_addresses)
+                    .await;
+            }
+        };
+
+        if misses.is_empty() {
```
dev-jodee, line 590: max_price_staleness_slots can make a cached price fail instead of refetch. Stale cache hits should probably be treated as misses.

crates/lib/src/cache.rs
```diff
@@ -343,6 +367,239 @@ impl CacheUtil {
 
         Ok(())
     }
+
+    fn get_price_key(mint_address: &str) -> String {
+        format!("{PRICE_CACHE_KEY_PREFIX}:{mint_address}")
```
dev-jodee: Price cache keys should include price_source, or source-mismatched cached values should be ignored. Right now Mock/Jupiter entries can overlap.

dev-jodee:
@greptile review

dev-jodee (approved):

PR #529, opened 2026-05-22
fix: cache payment-side token prices and add oracle HTTP timeouts

### LINKED ISSUE #519: Cache payment-side token prices/decimals and add oracle timeouts
I noticed something while looking at the payment estimation flow sometimes the same token's decimals and price get fetched more than once within a single request. The SPL outflow side already does a nice batch fetch by collecting distinct mints first, but the payment path does per-transfer lookups. Also the Jupiter oracle client has no HTTP timeouts set, so a hanging request just blocks and retries never kick in.

My thought was to bring the payment side in line with the SPL pattern pre-scan distinct mints, batch fetch prices/decimals once, then reuse. And add explicit connect + read timeouts (maybe 5-10s) to the oracle `reqwest::Client` so the retry logic actually gets a chance to do its thing.

Wanted to run this by you before touching anything. What do you think? If this sounds good, let's discuss or if you have other suggestions.

raushan728:
Added 5s connect and 10s read timeouts to the Jupiter oracle reqwest client via named Duration constants so retry logic can actually fire on hanging requests. Hanging server helper and HangingOracle extracted to oracle_mock.rs.

Refactored `calculate_payment_lamport_totals` to batch fetch prices via `CacheUtil::get_or_fetch_token_prices` and decimals once per unique mint instead of per transfer matching `calculate_spl_transfers_value_in_lamports`. Extracted repeated lamport math into `calculate_token_value_in_lamports_from_price` and updated `calculate_token_value_in_lamports` to use it.

Closes #519

### DIFF
--- a/crates/lib/src/oracle/oracle.rs
+++ b/crates/lib/src/oracle/oracle.rs
@@ -56,20 +56,12 @@ pub struct RetryingPriceOracle {
     oracle: Arc<dyn PriceOracle + Send + Sync>,
 }
 
-#[cfg(test)]
-thread_local! {
-    pub static TEST_ORACLE: std::cell::RefCell<Option<Arc<dyn PriceOracle + Send + Sync>>> = std::cell::RefCell::new(None);
-}
+const ORACLE_CONNECT_TIMEOUT_SECS: u64 = 5;
+const ORACLE_REQUEST_TIMEOUT_SECS: u64 = 10;
 
 pub fn get_price_oracle(
     source: PriceSource,
 ) -> Result<Arc<dyn PriceOracle + Send + Sync>, KoraError> {
-    #[cfg(test)]
-    {
-        if let Some(oracle) = TEST_ORACLE.with(|o| o.borrow().clone()) {
-            return Ok(oracle);
-        }
-    }
     match source {
         PriceSource::Jupiter => Ok(Arc::new(JupiterPriceOracle::new()?)),
         PriceSource::Mock => Ok(OracleUtil::get_mock_oracle_price()),
@@ -83,8 +75,8 @@ impl RetryingPriceOracle {
         oracle: Arc<dyn PriceOracle + Send + Sync>,
     ) -> Self {
         let client = Client::builder()
-            .connect_timeout(Duration::from_secs(5))
-            .timeout(Duration::from_secs(10))
+            .connect_timeout(Duration::from_secs(ORACLE_CONNECT_TIMEOUT_SECS))
+            .timeout(Duration::from_secs(ORACLE_REQUEST_TIMEOUT_SECS))
             .build()
             .expect("Failed to build reqwest client");
         Self { client, max_retries, base_delay, oracle }
--- a/crates/lib/src/oracle/oracle.rs
+++ b/crates/lib/src/oracle/oracle.rs
@@ -56,8 +56,8 @@ pub struct RetryingPriceOracle {
     oracle: Arc<dyn PriceOracle + Send + Sync>,
 }
 
-const ORACLE_CONNECT_TIMEOUT_SECS: u64 = 5;
-const ORACLE_REQUEST_TIMEOUT_SECS: u64 = 10;
+const ORACLE_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const ORACLE_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
 
 pub fn get_price_oracle(
     source: PriceSource,
@@ -75,8 +75,8 @@ impl RetryingPriceOracle {
         oracle: Arc<dyn PriceOracle + Send + Sync>,
     ) -> Self {
         let client = Client::builder()
-            .connect_timeout(Duration::from_secs(ORACLE_CONNECT_TIMEOUT_SECS))
-            .timeout(Duration::from_secs(ORACLE_REQUEST_TIMEOUT_SECS))
+            .connect_timeout(ORACLE_CONNECT_TIMEOUT)
+            .timeout(ORACLE_REQUEST_TIMEOUT)
             .build()
             .expect("Failed to build reqwest client");
         Self { client, max_retries, base_delay, oracle }
@@ -126,6 +126,7 @@ impl RetryingPriceOracle {
 mod tests {
 
     use super::*;
+    use crate::tests::oracle_mock::{spawn_hanging_server, HangingOracle};
     use std::time::Duration;
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


crates/lib/src/oracle/oracle.rs
```diff
@@ -46,9 +46,20 @@ pub struct RetryingPriceOracle {
...
+}
+
 pub fn get_price_oracle(
     source: PriceSource,
 ) -> Result<Arc<dyn PriceOracle + Send + Sync>, KoraError> {
+    #[cfg(test)]
+    {
+        if let Some(oracle) = TEST_ORACLE.with(|o| o.borrow().clone()) {
+            return Ok(oracle);
+        }
```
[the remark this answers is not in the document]
raushan728: All existing tests use the default `current_thread` runtime so this works correctly today. A follow-up can migrate to `OnceLock` or `Arc<Mutex<>>` if multi-threaded test cases are added

raushan728:
Ping! Just keeping this PR active.

crates/lib/src/oracle/oracle.rs
```diff
@@ -71,7 +72,12 @@ impl RetryingPriceOracle {
         base_delay: Duration,
         oracle: Arc<dyn PriceOracle + Send + Sync>,
     ) -> Self {
-        Self { client: Client::new(), max_retries, base_delay, oracle }
+        let client = Client::builder()
+            .connect_timeout(Duration::from_secs(5))
```
dev-jodee: constant should be at the top

crates/lib/src/token/token.rs
```diff
@@ -916,35 +932,115 @@ impl TokenUtil {
...
+                current_slot,
+            )
+            .await?;
+        }
 
-                totals.checked_add_assign(PaymentLamportTotals {
-                    inflow: inflow_lamports,
-                    outflow: outflow_lamports,
+        for (is_inflow, is_outflow, token_mint, inflow_amount, amount) in valid_transfers {
+            let decimals = Self::get_mint_decimals(config, rpc_client, &token_mint).await?;
```
dev-jodee: decimals can be fetched per mint instead of per transfer

crates/lib/src/token/token.rs
```diff
@@ -916,35 +932,115 @@ impl TokenUtil {
...
+                &format!(" for {mint_addr}"),
+                current_slot,
+            )
+            .await?;
+        }
 
-                totals.checked_add_assign(PaymentLamportTotals {
-                    inflow: inflow_lamports,
-                    outflow: outflow_lamports,
+        for (is_inflow, is_outflow, token_mint, inflow_amount, amount) in valid_transfers {
```
dev-jodee: lets use a struct instead of a 5 member tuple

crates/lib/src/token/token.rs
```diff
@@ -916,35 +932,115 @@ impl TokenUtil {
...
+            let decimals = Self::get_mint_decimals(config, rpc_client, &token_mint).await?;
+            let price = prices.get(&token_mint.to_string()).ok_or_else(|| {
+                KoraError::RpcError(format!("No price data for mint {token_mint}"))
+            })?;
+
+            let inflow_lamports = if is_inflow {
+                let amount_decimal = Decimal::from_u64(inflow_amount).ok_or_else(|| {
```
dev-jodee: should this call Self::calculate_token_value_in_lamports_from_price instead of duplicating the math?

crates/lib/src/token/token.rs
```diff
@@ -916,35 +932,115 @@ impl TokenUtil {
...
+
+                lamports_decimal.floor().to_u64().ok_or_else(|| {
+                    KoraError::ValidationError("Lamports value overflow".to_string())
+                })?
+            } else {
+                0
+            };
+
+            let outflow_lamports = if is_outflow {
+                let amount_decimal = Decimal::from_u64(amount).ok_or_else(|| {
```
dev-jodee: should this call Self::calculate_token_value_in_lamports_from_price instead of duplicating the math?

crates/lib/src/oracle/oracle.rs
```diff
@@ -56,9 +46,20 @@ pub struct RetryingPriceOracle {
...
 
+#[cfg(test)]
+thread_local! {
+    pub static TEST_ORACLE: std::cell::RefCell<Option<Arc<dyn PriceOracle + Send + Sync>>> = std::cell::RefCell::new(None);
+}
+
 pub fn get_price_oracle(
     source: PriceSource,
 ) -> Result<Arc<dyn PriceOracle + Send + Sync>, KoraError> {
+    #[cfg(test)]
```
dev-jodee: we don't want to use cfg test flag inside functions directly, just like we do with the cache, you should overwrite at the import level

```#[cfg(not(test))]
  use crate::oracle::get_price_oracle;

  #[cfg(test)]
  use crate::tests::oracle_mock::get_price_oracle;```

**Review by dev-jodee — CHANGES_REQUESTED** [review]

File: crates/lib/src/oracle/oracle.rs [hunk 8]
```diff
@@ -71,7 +74,12 @@ impl RetryingPriceOracle {
         base_delay: Duration,
         oracle: Arc<dyn PriceOracle + Send + Sync>,
     ) -> Self {
-        Self { client: Client::new(), max_retries, base_delay, oracle }
+        let client = Client::builder()
+            .connect_timeout(Duration::from_secs(ORACLE_CONNECT_TIMEOUT_SECS))
```
dev-jodee: Can we make the const be the actual Duration::from, so we don't initialize it every call, just once

crates/lib/src/oracle/oracle.rs
```diff
@@ -209,4 +217,60 @@ mod tests {
...
+        let server_url = format!("http://127.0.0.1:{}", port);
+
+        let _server_handle = std::thread::spawn(move || {
+            if let Ok((stream, _)) = listener.accept() {
+                std::thread::sleep(Duration::from_secs(30));
+                let _ = stream;
+            }
+        });
+
+        struct HangingOracle {
```
dev-jodee: All these oracle mock related struct and functions should be in the oracle_mock.rs file

dev-jodee: same with the listener above, could have like `spawn_hanging_server`

crates/lib/src/oracle/oracle.rs
```diff
@@ -209,4 +217,60 @@ mod tests {
...
+                Ok(HashMap::new())
+            }
+        }
+
+        let hanging_oracle = Arc::new(HangingOracle { url: server_url });
+        let retrying_oracle =
+            RetryingPriceOracle::new(1, Duration::from_millis(10), hanging_oracle);
+
+        let result = tokio::time::timeout(
+            Duration::from_secs(15),
```
dev-jodee, line 231: Think this doesn't make sense, if 15 is max outer, and you have retrying oracle max retry 1 and 10 ms, 10 + 10 = 20, so outer loop will fail before?

crates/lib/src/token/token.rs
```diff
@@ -431,6 +431,36 @@ impl TokenUtil {
         Ok((token_price, decimals))
     }
 
+    fn calculate_token_value_in_lamports_from_price(
```
dev-jodee, line 434: If we're going to have this function that's a close duplicate of pub async fn calculate_token_value_in_lamports, we should change the existing one to use this new one 

```
(
      amount: u64,
      mint: &Pubkey,                                                                                                   
      rpc_client: &RpcClient,
      config: &Config,
  ) -> Result<u64, KoraError> {
      let (token_price, decimals) =
          Self::get_token_price_and_decimals(mint, rpc_client, config).await?;
      Self::calculate_token_value_in_lamports_from_price(amount, token_price.price, decimals)
  }```

File: crates/lib/src/token/token.rs [hunk 12]
```diff
@@ -2097,4 +2187,78 @@ mod tests_token {
...
+            &mut transaction_resolved,
+            &rpc_client,
+            &expected_destination_owner,
+            None,
+        )
+        .await;
+
+        assert!(result.is_ok());
+        let totals = result.unwrap();
+        assert_eq!(totals.inflow, 22_500_000);
```
dev-jodee: shouldn't we still assert ```mock_oracle.expect_get_prices().times(1)``` thats the point of the test right ?

dev-jodee (changes requested):
Some other changes to do, also can you not rebase / force push your changes and just commit on top of whats already there? Makes it easier to review just the diff from last reviewed commit. Thanks !

crates/lib/src/oracle/oracle.rs
raushan728: The `10` in `RetryingPriceOracle::new(1, Duration::from_millis(10), ...)` is the retry backoff delay (10ms), not the request timeout. The reqwest client has a 10s read timeout. With 1 retry and 10ms backoff, total time is ~10s which is well within the 15s outer window.

raushan728:
> Some other changes to do, also can you not rebase / force push your changes and just commit on top of whats already there? Makes it easier to review just the diff from last reviewed commit. Thanks !

I thought rebasing was the normal flow for keeping a PR up to date, so that's why I did it. Anyway, noted for future Kora contributions.

dev-jodee (approved):

PR #534, opened 2026-05-28
feat(devnet-deploy-paymaster): add inactive-program reaper

dev-jodee:
## Summary
- New `devnet_deploy_reaper` binary in `examples/devnet-deploy-paymaster/` that scans the chain for programs whose upgrade authority is the paymaster's fee payer, classifies activity via `getSignaturesForAddress` (with a `getBlockTime(last_state_slot)` fallback), and closes idle ones — recovering rent to the fee payer. Covers BPF Loader Upgradeable (v3) and Loader-v4.
- Example crate restructured: renamed `devnet-deploy-paymaster`, added `[lib]` + second `[[bin]] devnet_deploy_reaper`. Production `crates/lib` and `crates/cli` untouched — keeps devnet-only logic colocated with devnet-only config.
- `Dockerfile.reaper` (multi-stage local-source build) + `cloudbuild.reaper.yaml` (workspace-context build).
- `.github/workflows/deploy-devnet-paymaster.yml` now also builds the reaper image and runs `gcloud run jobs update` against `$CLOUD_RUN_REAPER_JOB`. New Doppler key required.

## How it works
1. **Discovery** — for v3, one `getProgramAccounts(BPF_LOADER_UPGRADEABLE)` filtered by upgrade authority returns the owned `ProgramData` accounts, then one narrow `getProgramAccounts` per `ProgramData` resolves its sibling `Program` pointer. For v4, one `getProgramAccounts(LoaderV4)` filtered by authority at `offset_of!(LoaderV4State, authority_address_or_next_version)`.
2. **Activity** — newest signature's `block_time` vs `now − threshold`; fallback to `getBlockTime(last_state_slot)` when no sigs; default idle.
3. **Close** — v3: `close_any(program_data, fee_payer, fee_payer, program)`. v4: `Retract` + `SetProgramLength(0, recipient=fee_payer)` in one tx. Signed locally with the same `Arc<Signer>` resolved at startup, threaded through `ReaperConfig` (no per-call pool re-selection).

CLI flags: `--threshold-hours <u64>` (default `168`), `--dry-run`, `--max-closes <n>`. Always processes both v3 and v4.

## Test Plan
- [x] `cargo test -p devnet-deploy-paymaster --lib` (5 unit tests covering slot parsing + v3/v4 close-instruction shape)
- [x] `cargo fmt` clean
- [x] `cargo clippy -p devnet-deploy-paymaster --all-targets -- -D warnings` clean
- [ ] After merge: create Cloud Run Job + Cloud Scheduler entry per README, then trigger manually with `--dry-run` and `--max-closes=5` before letting the daily cron run free

## Out of scope (open follow-ups)
- Redis audit trail — deferred; Cloud Logging + on-chain sigs are sufficient
- Integration test against a local validator (deploy → reap → assert rent reclaimed)
- A `getReapableProgramsForOwner` RPC method so users can preview before their program gets reaped

## Required infra (one-time, documented in README)
- Create Cloud Run Job (`CLOUD_RUN_REAPER_JOB`) with the KMS-signer SA
- Create Cloud Scheduler entry → invoke the Job daily (suggested `0 3 * * *`)
- Add Doppler key `CLOUD_RUN_REAPER_JOB`

## 📊 Unit Test Coverage
Coverage

**Unit Test Coverage: 85.1%**

[View Detailed Coverage Report](https://github.com/solana-foundation/kora/actions/runs/26586065320)


examples/devnet-deploy-paymaster/src/reaper/closer.rs
```diff
@@ -0,0 +1,148 @@
...
+        Some(fee_payer),
+        Some(&program.program),
+    )])
+}
+
+fn build_v4_close(fee_payer: &Pubkey, program: &OwnedProgram) -> Vec<Instruction> {
+    vec![
+        loader_v4::retract(&program.program, fee_payer),
+        loader_v4::set_program_length(&program.program, fee_payer, 0, fee_payer),
+    ]
```
raushan728, line 77: If a program is already in a `retracted` state, the `loader_v4::retract` instruction will fail and revert the entire transaction. We should check the program state first or make this idempotent so we don't get stuck with 'Failed' reports on every run for the same program

examples/devnet-deploy-paymaster/src/reaper/closer.rs
```diff
@@ -0,0 +1,148 @@
...
+    ));
+    let mut tx = TransactionUtil::new_unsigned_versioned_transaction(message);
+    let signature = signer
+        .sign_message(&tx.message.serialize())
+        .await
+        .map_err(|e| anyhow!("sign_message: {e}"))?;
+    tx.signatures = vec![signature];
+
+    let sig = rpc
+        .send_and_confirm_transaction_with_spinner(&tx)
```
raushan728, line 49: Since this runs as a Cloud Run Job, using `with_spinner` will create a lot of noise in the logs with terminal escape characters. It's better to use the standard `send_and_confirm_transaction` for cleaner logging

.github/workflows/deploy-devnet-paymaster.yml
```diff
@@ -92,3 +92,52 @@ jobs:
 
       - name: Print service URL
         run: echo "Deployed ${{ env.CLOUD_RUN_SERVICE }} -> ${{ steps.deploy.outputs.url }}"
+
+      - id: build-reaper
```
raushan728, line 96: I think we should add a guard here like `if: env.CLOUD_RUN_REAPER_JOB != ''` if the reaper infra isn't provisioned yet, this workflow will still submit a build and poll for 40 minutes before failing, which will mark the entire deploy red unnecessarily

raushan728 (commented):
PR is already merged and I might be mistaken, but I noticed a few things.

PR #538, opened 2026-05-29
docs(devnet-deploy-paymaster): user-facing usage + cleaner env-var docs

dev-jodee:
## Summary
- Add **Using the paymaster** section at the top of the README — endpoint, constraints (paymaster owns upgrade authority, programs idle >7d get reaped), reference flow pointing at `src/main.rs` + `smoke-test.sh`, JSON-RPC call shape.
- Demote operator content under **Deploying your own paymaster**.
- Replace **Doppler config** with **Environment variables** — drop Doppler-specific narration, just list the vars the workflow needs. Operators can plug in whatever secret store they use.

## Test Plan
- [x] Render README on the PR; sections flow user-first → operator-first
- [ ] After merge, double-check the smoke-test.sh command in the new section actually runs against the live deployer

### DIFF
--- a/examples/devnet-deploy-paymaster/README.md
+++ b/examples/devnet-deploy-paymaster/README.md
@@ -1,27 +1,66 @@
 # Devnet Deploy Paymaster
 
-Reference Kora paymaster that sponsors program deploys on Solana devnet, with
-loader-v3 / loader-v4 upgrade authority pinned to Kora so users can't drain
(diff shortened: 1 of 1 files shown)


dev-jodee:
Superseded by #539 — the new `kora-deploy` crate makes the inline Rust example obsolete. README changes will land in #539 directly.

PR #556, opened 2026-06-11
Add async transaction sending

n00m4d:
# What
Adds an optional `confirmation` parameter to `signAndSendTransaction`, letting callers choose how long the server waits on the broadcast before responding:

- `confirmed` (default) — waits for on-chain confirmation
- `sent` — returns once the RPC node accepts the transaction
- `none` — broadcasts in the background and returns as soon as signing completes

# Why
The previous behavior always waited for full on-chain confirmation, adding latency for callers that don't need it. The faster modes let clients trade confirmation guarantees for lower latency.

# How
- New `ConfirmationMode` enum threaded through `sign_and_send_transaction`, defaulting to `Confirmed` for backward compatibility.
- Fast modes (`sent`, `none`) skip preflight since validation already simulates the transaction.
- `none` returns the transaction's first signature immediately and broadcasts via `tokio::spawn`; broadcast failures are logged server-side rather than returned. It rejects transactions missing required signatures up front to avoid silent background failures.
- Exposed in the OpenAPI docs and the TypeScript SDK (`ConfirmationMode` type), with unit and integration test coverage.

### DIFF
--- a/crates/lib/src/transaction/versioned_transaction.rs
+++ b/crates/lib/src/transaction/versioned_transaction.rs
@@ -560,6 +560,17 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
         // transiently even though the transaction is valid.
         let send_config = RpcSendTransactionConfig { skip_preflight: true, ..Default::default() };
 
+        // Skipping preflight also skips the RPC node's signature verification, so a
+        // transaction with an unfilled co-signer slot (possible when validation ran
+        // with sig_verify off) would be accepted by the RPC but dropped silently by
+        // validators. Reject it here so the caller gets an error instead of a
+        // signature for a transaction that can never land.
+        if transaction.signatures.iter().any(|s| *s == Signature::default()) {
+            return Err(KoraError::InvalidTransaction(
+                "Transaction is missing required signatures".to_string(),
+            ));
+        }
+
         match confirmation {
             ConfirmationMode::Confirmed => {
                 let signature = rpc_client
@@ -578,16 +589,6 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
                 Ok((signature.to_string(), encoded))
             }
             ConfirmationMode::None => {
-                // The response is built before the broadcast runs, so a transaction
-                // that can never land (missing co-signer signature when sig_verify is
-                // off) must be rejected here instead of failing silently in the
-                // background.
-                if transaction.signatures.iter().any(|s| *s == Signature::default()) {
-                    return Err(KoraError::InvalidTransaction(
-                        "Transaction is missing required signatures".to_string(),
-                    ));
-                }
-
                 // A Solana transaction is identified by its first signature, which is
                 // already present once signing completes — so the caller gets it
                 // without waiting for the broadcast.
--- a/crates/lib/src/transaction/versioned_transaction.rs
+++ b/crates/lib/src/transaction/versioned_transaction.rs
@@ -91,20 +91,21 @@ impl Deref for VersionedTransactionResolved {
     }
 }
 
-/// How long `signAndSendTransaction` waits on the broadcast before responding.
+/// The transaction lifecycle milestone `signAndSendTransaction` waits for before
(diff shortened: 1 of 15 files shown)


amilz:
Cool idea/thanks for the PR. One immediate thought `confirmation` is a loaded term so I'm wondering if we us a different name like `broadcast_trigger` or something like that? WDYT @n00m4d @dev-jodee

tests/rpc/transaction_signing.rs
```diff
@@ -147,6 +147,96 @@ async fn test_sign_and_send_transaction_legacy() {
...
+    let response = result.unwrap();
+
+    // The signature is derived from the signed transaction before the background
+    // broadcast runs, so it must match the transaction's first signature.
+    let signature = response["signature"].as_str().expect("Expected signature in response");
+    let signed_transaction =
+        response["signed_transaction"].as_str().expect("Expected signed_transaction in response");
+    let decoded = TransactionUtil::decode_b64_transaction(signed_transaction)
+        .expect("Returned signed transaction should decode");
+    assert_eq!(
```
dev-jodee, line 274: Could this poll getSignatureStatuses for a few seconds? Because now a broken bg spawn would still succeed for this test right?

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -523,19 +548,78 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
...
         let (transaction, encoded) =
             self.sign_transaction(config, signer, rpc_client, true).await?;
 
-        // Send and confirm transaction
-        let signature = rpc_client
-            .send_and_confirm_transaction(&transaction)
-            .await
-            .map_err(|e| KoraError::RpcError(sanitize_error!(e)))?;
+        // Validation already simulated the transaction, so the fast modes skip
+        // preflight: a second simulation only delays landing and can fail
```
dev-jodee, line 560: nit: the validation simulation runs before Kora signs and with the caller's sig_verify (default false), so we should add to the comment that it doesn't cover signature validity, just to be 100% clear

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -523,19 +548,78 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
...
+                            "Signed transaction has no signatures".to_string(),
+                        )
+                    })?
+                    .to_string();
+
+                // Broadcast in the background so the response returns instantly. The
+                // send is fire-and-forget: failures are logged rather than returned to
+                // the caller, who can rebroadcast the returned signed transaction.
+                let rpc_client = std::sync::Arc::clone(rpc_client);
+                let log_signature = signature.clone();
```
dev-jodee, line 625: nit: just need to make it crystal clear that None will also fail on potential transient rpc errors, etc. so there's really 0 delivery guarantee

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -523,19 +548,78 @@ impl VersionedTransactionOps for VersionedTransactionResolved {
...
 
-        // Send and confirm transaction
-        let signature = rpc_client
-            .send_and_confirm_transaction(&transaction)
-            .await
-            .map_err(|e| KoraError::RpcError(sanitize_error!(e)))?;
+        // Validation already simulated the transaction, so the fast modes skip
+        // preflight: a second simulation only delays landing and can fail
+        // transiently even though the transaction is valid.
+        let send_config = RpcSendTransactionConfig { skip_preflight: true, ..Default::default() };
```
dev-jodee: nit: rename to something that explains it's only for "Sent" and "None" mode, as Confirmed mode odesn't use it in the match below

dev-jodee (commented):
I like the idea, some comments of my own, and I agree with @amilz on the word confirmation, and I think the ai review has some good comments as well :)

dev-jodee (dismissed):
looks good, can you fix CI ?

dev-jodee (approved):

PR #559, opened 2026-06-13
feat: implement identity-aware rate limiting and configurable CORS

raushan728:
`rpc_server`: replace global rate limiter with identity-aware limiting and configurable CORS

- Add `global_rate_limit` (pre-auth) to throttle unauthenticated spam cheaply
- Replace shared `RateLimitLayer` with per-identity token buckets (`IdentityRateLimitLayer`)
- Add `cors_allow_origins` config field, replacing hardcoded `allow_origin(Any)`
- Migrate `api_key` -> `api_keys: Vec<String>` with backward-compatible deserialization
- Add `http_rejections_total` Prometheus counter distinguishing `auth_failure` vs `rate_limit`
- Reorder middleware: `metrics` -> `CORS` -> `global limit` -> `auth` -> `per-identity limit` -> `recaptcha`
- Update all `kora.toml` examples with new fields and deprecation notes

### DIFF
--- a/crates/lib/src/rpc_server/auth.rs
+++ b/crates/lib/src/rpc_server/auth.rs
@@ -10,27 +10,45 @@ use jsonrpsee::server::logger::Body;
 use sha2::Sha256;
 use subtle::ConstantTimeEq;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ClientIdentity(pub String);
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum RejectionReason {
+    AuthFailure,
+    RateLimit,
+}
+
+impl RejectionReason {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            RejectionReason::AuthFailure => "auth_failure",
+            RejectionReason::RateLimit => "rate_limit",
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct ApiKeyAuthLayer {
-    api_key: String,
+    api_keys: Vec<String>,
 }
 
 impl ApiKeyAuthLayer {
-    pub fn new(api_key: String) -> Self {
-        Self { api_key }
+    pub fn new(api_keys: Vec<String>) -> Self {
+        Self { api_keys }
     }
 }
 
 #[derive(Clone)]
 pub struct ApiKeyAuthService<S> {
     inner: S,
-    api_key: String,
+    api_keys: Vec<String>,
 }
 
 impl<S> tower::Layer<S> for ApiKeyAuthLayer {
     type Service = ApiKeyAuthService<S>;
     fn layer(&self, inner: S) -> Self::Service {
-        ApiKeyAuthService { inner, api_key: self.api_key.clone() }
+        ApiKeyAuthService { inner, api_keys: self.api_keys.clone() }
     }
 }
 
@@ -53,12 +71,13 @@ where
     }
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
-        let api_key = self.api_key.clone();
+        let api_keys = self.api_keys.clone();
         let mut inner = self.inner.clone();
 
         Box::pin(async move {
-            let unauthorized_response =
+            let mut unauthorized_response =
                 build_response_with_graceful_error(None, StatusCode::UNAUTHORIZED, "");
+            unauthorized_response.extensions_mut().insert(RejectionReason::AuthFailure);
 
             let (parts, body_bytes) = extract_parts_and_body_bytes(request).await;
 
@@ -72,10 +91,22 @@ where
             }
 
             // Check for API key header
-            let req = Request::from_parts(parts, Body::from(body_bytes));
-            if let Some(provided_key) = req.headers().get(X_API_KEY) {
+            let mut req = Request::from_parts(parts, Body::from(body_bytes));
+            let provided_key = req.headers().get(X_API_KEY).cloned();
+            if let Some(provided_key) = provided_key {
                 // Constant-time comparison prevents timing attacks
-                if provided_key.as_bytes().ct_eq(api_key.as_bytes()).into() {
+                let mut matched = false;
+                for key in &api_keys {
+                    if provided_key.as_bytes().ct_eq(key.as_bytes()).into() {
+                        matched = true;
+                        break;
+                    }
+                }
+
+                if matched {
+                    if let Ok(key_str) = provided_key.to_str() {
+                        req.extensions_mut().insert(ClientIdentity(format!("apikey:{}", key_str)));
+                    }
                     return inner.call(req).await;
                 }
             }
@@ -140,8 +171,9 @@ where
         let mut inner = self.inner.clone();
 
         Box::pin(async move {
-            let unauthorized_response =
+            let mut unauthorized_response =
                 build_response_with_graceful_error(None, StatusCode::UNAUTHORIZED, "");
+            unauthorized_response.extensions_mut().insert(RejectionReason::AuthFailure);
 
             let signature_header = request.headers().get(X_HMAC_SIGNATURE).cloned();
             let timestamp_header = request.headers().get(X_TIMESTAMP).cloned();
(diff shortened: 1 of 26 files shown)


crates/lib/src/rpc_server/rate_limit.rs
```diff
@@ -0,0 +1,192 @@
...
+            None => UNAUTHENTICATED_IDENTITY.to_string(),
+        };
+
+        let allowed = {
+            let mut map = self.state.write();
+            let bucket = map.entry(identity).or_insert_with(|| {
+                TokenBucket::new(self.rate_limit as f64, self.rate_limit as f64)
+            });
+            bucket.consume(1.0)
+        };
```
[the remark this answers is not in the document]
raushan728: Map keys come from ClientIdentity, which is only apikey:<configured-key>, hmac:global, or unauthenticated all static, config-derived, not request-driven. So size is bounded by (num configured API keys) + 2, doesn't grow with traffic. Added a comment in code calling this out; eviction felt like overkill for this bound.

crates/lib/src/rpc_server/server.rs
```diff
@@ -49,9 +49,34 @@ pub async fn run_rpc_server(rpc: KoraRpc, port: u16) -> Result<ServerHandles, an
         return Err(anyhow::anyhow!("Usage limiter initialization failed: {e}"));
     }
 
+    let config = get_config()?;
```
dev-jodee: we should have this in its own function, to keep this clean

crates/lib/src/rpc_server/server.rs
```diff
@@ -83,29 +106,41 @@ pub async fn run_rpc_server(rpc: KoraRpc, port: u16) -> Result<ServerHandles, an
...
-                .map(ApiKeyAuthLayer::new),
-        )
+        .option_layer(api_keys_config.map(ApiKeyAuthLayer::new))
         // Add authentication layer for HMAC if configured
         .option_layer(
             get_value_by_priority("KORA_HMAC_SECRET", config.kora.auth.hmac_secret.clone())
                 .map(|secret| HmacAuthLayer::new(secret, config.kora.auth.max_timestamp_age)),
         )
+        // Identity-aware rate limiting
+        .layer(IdentityRateLimitLayer::new(config.kora.rate_limit))
```
dev-jodee, line 212: this is done after the auth now, so request that fail the auth part won't be limited. So unauthenticated traffic is now unthrottled. Think we still need global ip limiter before auth, as auth is a pretty heavy workflow, so someone could spam an endpoint and overload the server trying to do the auth over and over

crates/lib/src/rpc_server/auth.rs
```diff
@@ -219,7 +264,9 @@ where
 
             // Reconstruct the request with the consumed body
             let new_body = Body::from(body_bytes);
-            let new_request = Request::from_parts(parts, new_body);
+            let mut new_request = Request::from_parts(parts, new_body);
+            // HMAC currently uses a single global shared secret, so all HMAC clients share one rate-limit bucket
+            new_request.extensions_mut().insert(ClientIdentity("hmac:global".to_string()));
```
dev-jodee: I think this overwrite the api key ClientIdentity you created in the case where both api key and hmac are enabled. ClientIdentity is a map based on the Type (both strings here)

crates/lib/src/rpc_server/auth.rs
```diff
@@ -72,10 +100,20 @@ where
...
-                if provided_key.as_bytes().ct_eq(api_key.as_bytes()).into() {
+                let mut matched = false;
+                for key in &api_keys {
+                    let is_match: bool = provided_key.as_bytes().ct_eq(key.as_bytes()).into();
+                    matched |= is_match;
+                }
+
+                if matched {
+                    if let Ok(key_str) = provided_key.to_str() {
+                        req.extensions_mut().insert(ClientIdentity(format!("apikey:{}", key_str)));
```
dev-jodee: not a fan of storing plain text api key as the hashmap key in the limiter, can probably just store its index?

crates/lib/src/rpc_server/rate_limit.rs
```diff
@@ -0,0 +1,193 @@
...
+        Box<dyn std::future::Future<Output = Result<Self::Response, Self::Error>> + Send>,
+    >;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        // rate_limit=0 disables per-identity limiting
+        if self.rate_limit == 0 {
```
dev-jodee: Think we should use an option here instead of a sentinel value, before 0 would panic in Tower's rate limit, now it would just mean no rate limit

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -330,9 +330,18 @@ impl UsageTracker {
         let usage_config = &config.kora.usage_limit;
 
         let set_limiter = |limiter| {
-            USAGE_LIMITER.set(limiter).map_err(|_| {
-                KoraError::InternalServerError("Usage limiter already initialized".to_string())
-            })
+            if USAGE_LIMITER.set(limiter).is_err() {
+                #[cfg(test)]
```
dev-jodee: can't have cfg test flags in the production code, think I mentioned it before, this should be done like we do cache mocking i.e.

dev-jodee (changes requested):

crates/lib/src/rpc_server/server.rs
raushan728: Extracted into build_allow_origin() helper.

crates/lib/src/rpc_server/server.rs
raushan728: Added `global_rate_limit: Option<u64>` config field. Pre-auth IdentityRateLimitLayer placed before auth layers at that point no ClientIdentity exists so all requests share the "unauthenticated" bucket, acting as a global guard. `tower::limit::RateLimitLayer` couldn't be used as jsonrpsee requires clone on middleware stack.

crates/lib/src/rpc_server/auth.rs
raushan728: HmacAuthService now checks if ClientIdentity is already set before inserting "hmac:global".

crates/lib/src/rpc_server/rate_limit.rs
raushan728: `rate_limit: Option<u64>`, None disables limiting, Some(n) enforces n req/s per identity.

raushan728:
Because there was conflict here, I had to rebase. From ba3bd1e7232986e38498d5dbafa537f0260fc68c, there are review addresses.

dev-jodee:
last thing I'd say is to update the kora.toml(s) that we have with the new values so the example shows all thats available

dev-jodee (changes requested):
comment I mentioned above^

raushan728:
> last thing I'd say is to update the kora.toml(s) that we have with the new values so the example shows all thats available

Done all kora.toml example files updated. Addresses in 7bbb31b4c265c85e39ee789bb749ea4cc4a98a4e

raushan728:
After merging #609 a conflict came up in this PR. I just clicked on `accept both changes` because only the tests had an issue, and the author of #609 hadn't added a closing brace so I added that. This just made me think that the tests aren't really organized they're just scattered around randomly. In my opinion, everything should be properly organized so that it helps when writing similar tests later on. I'm not sure if this makes sense or not.

dev-jodee:
Hey @raushan728 the pr is in the right direction but we need to split it, it's a bit too big and does too many different things, I think the following prs split make sense

1. CORS
2. Many api keys support
3. Prometheus metric
4. global_rate_limit 

I think the more advanced rate limiter we can revisit but for now the 4 above would be good as split prs, and easier for me to review

raushan728:
Hey, I'll open 4 PRs. I'll leave this PR open as a reference(draft) and cherry-pick the relevant changes into each new branch. Once all 4 land, I'll rebase this PR on top with just the per-identity IdentityRateLimitLayer changes (the advanced rate limiter for issue 5).

Starting with CORS first.

raushan728:
I'll have to do double work here work that had already been done. 😩

PR #565, opened 2026-06-16
feat: batch-fetch and cache address lookup tables for V0 transactions

raushan728:
V0 transactions resolved ALTs one at a time via `CacheUtil::get_account`, with a TODO comment noting caching was never added. Bundles re-resolved the same ALT for every transaction that referenced it.
 
- Added `CacheUtil::get_multiple_accounts` cache-aware batch fetch using Redis `mget` + RPC `get_multiple_accounts` for misses
- `resolve_lookup_table_addresses` now batches all ALT pubkeys into one call instead of looping; both paths deduplicate via `HashSet` before fetching
- Bundle processing shares one `AltCache` (`HashMap<Pubkey, Vec<Pubkey>>`) across all transactions in the same bundle repeated ALT fetched once per bundle, not once per tx. Single-transaction RPC handlers unaffected (pass `None`)
- Deduped miss pubkeys before batch fetch (same ALT referenced twice in one tx)
- Cache partial results before returning error successfully fetched accounts are written to Redis even if one account in the batch is missing

### DIFF
--- a/crates/lib/src/cache.rs
+++ b/crates/lib/src/cache.rs
@@ -670,6 +670,144 @@ impl CacheUtil {
             ))
         })
     }
+
+    /// Get multiple accounts directly from RPC (bypassing cache)
+    async fn get_multiple_accounts_from_rpc(
+        rpc_client: &RpcClient,
+        pubkeys: &[Pubkey],
+    ) -> Result<Vec<Account>, KoraError> {
+        match rpc_client.get_multiple_accounts(pubkeys).await {
+            Ok(accounts_opt) => {
+                let mut result = Vec::with_capacity(pubkeys.len());
+                for (i, acc_opt) in accounts_opt.into_iter().enumerate() {
+                    match acc_opt {
+                        Some(acc) => result.push(acc),
+                        None => return Err(KoraError::AccountNotFound(pubkeys[i].to_string())),
+                    }
+                }
+                Ok(result)
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    /// Get multiple accounts, using Redis cache when available.
+    pub async fn get_multiple_accounts(
+        config: &Config,
+        rpc_client: &RpcClient,
+        pubkeys: &[Pubkey],
+    ) -> Result<Vec<Account>, KoraError> {
+        if pubkeys.is_empty() {
+            return Ok(vec![]);
+        }
+
+        // If cache is disabled, go directly to RPC
+        if !CacheUtil::is_cache_enabled(config) {
+            return Self::get_multiple_accounts_from_rpc(rpc_client, pubkeys).await;
+        }
+
+        let pool = match CACHE_POOL.get() {
+            Some(Some(pool)) => pool,
+            _ => return Self::get_multiple_accounts_from_rpc(rpc_client, pubkeys).await,
+        };
+
+        let mut conn = match Self::get_connection(pool).await {
+            Ok(c) => c,
+            Err(e) => {
+                log::warn!(
+                    "Failed to get cache connection, falling back to RPC: {}",
+                    sanitize_error!(e)
+                );
+                return Self::get_multiple_accounts_from_rpc(rpc_client, pubkeys).await;
+            }
+        };
+
+        let keys: Vec<String> = pubkeys.iter().map(Self::get_account_key).collect();
+        let raw: Vec<Option<String>> = match conn.mget(&keys).await {
+            Ok(r) => r,
+            Err(e) => {
+                log::warn!("Failed to get multiple accounts from cache: {}", sanitize_error!(e));
+                vec![None; pubkeys.len()]
+            }
+        };
+
+        let current_time = chrono::Utc::now().timestamp();
+        let mut results: Vec<Option<Account>> = vec![None; pubkeys.len()];
+        let mut misses: Vec<(usize, Pubkey)> = Vec::new();
+
+        for (i, (pubkey, cached_str_opt)) in pubkeys.iter().zip(raw.into_iter()).enumerate() {
+            let mut hit = false;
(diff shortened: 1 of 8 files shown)


crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -547,33 +549,108 @@ impl LookupTableUtil {
...
-                        KoraError::RpcError(format!(
-                            "Failed to fetch lookup table: {}",
+        if lookup_table_lookups.is_empty() {
+            return Ok(resolved_addresses);
+        }
+
+        let mut full_address_lists: Vec<Vec<Pubkey>> =
+            Vec::with_capacity(lookup_table_lookups.len());
+
+        if let Some(cache) = alt_cache {
```
dev-jodee: think this branching and the else branching are both very similar. Enough that we could make it into 1 instead. something like 

`
let mut local = HashMap::new(); 
let cache = alt_cache.unwrap_or(&mut local);
`

crates/lib/src/cache.rs
```diff
@@ -670,6 +670,144 @@ impl CacheUtil {
...
+            }
+
+            if has_pipe_ops {
+                if let Err(e) = pipe.query_async::<()>(&mut conn).await {
+                    log::warn!("Failed to cache accounts in batch: {}", sanitize_error!(e));
+                }
+            }
+        }
+
+        Ok(results.into_iter().flatten().collect())
```
dev-jodee: We should error if results has any None, flatten would "hide" the None value

crates/lib/src/token/token.rs
```diff
@@ -1197,7 +1197,7 @@ mod tests_token {
 
     #[tokio::test]
     async fn test_get_mint_account_not_found() {
-        let _lock = ConfigMockBuilder::new().build_and_setup();
+        let _lock = ConfigMockBuilder::new().with_cache_enabled(false).build_and_setup();
```
dev-jodee, line 1200: why was this needed in those tests? seems like token isn't affected?

crates/lib/src/cache.rs
```diff
@@ -670,6 +670,144 @@ impl CacheUtil {
...
+                            Err(e) => {
+                                log::warn!(
+                                    "Failed to serialize cache data for {}: {}",
+                                    pubkey,
+                                    sanitize_error!(e)
+                                );
+                            }
+                        }
+                    }
+                    None => return Err(KoraError::AccountNotFound(pubkey.to_string())),
```
dev-jodee: This will make it so none of the fetched accounts are cached, we should prob cache the ones that were successful even if only 1 fails ? That's how the old behavior was doing it

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -111,6 +111,7 @@ impl VersionedTransactionResolved {
         config: &Config,
         rpc_client: &RpcClient,
         sig_verify: bool,
+        alt_cache: Option<&mut HashMap<Pubkey, Vec<Pubkey>>>,
```
dev-jodee: nit: type alias so it's cleaner

dev-jodee (changes requested):

raushan728:
>  Note: dropped 2 tests for the Redis cache-hit path needed a live Redis instance, CI doesn't run one, and no existing test in `cache.rs` does either.
>
Restored the two Redis cache-hit tests `test_redis_get_multiple_accounts_all_cache_hits`, `test_redis_get_multiple_accounts_partial_cache_miss` since redis is being added to CI via a separate PR.

crates/lib/src/token/token.rs
raushan728: These token tests don't interact with cache. The explicit `with_cache_enabled(false)` guards them from the redis pool that will be active in CI once the Redis service lands without it they'd silently take the cache path instead of the mocked RPC path.

raushan728:
Review address form 3823d51667f1c0b6ed48cb8e0e56c04c441c873e

crates/lib/src/transaction/versioned_transaction.rs
```diff
@@ -649,33 +653,67 @@ impl LookupTableUtil {
...
-            // Parse the lookup table account data to get the actual addresses
-            let address_lookup_table = AddressLookupTable::deserialize(&lookup_table_account.data)
-                .map_err(|e| {
-                    KoraError::InvalidTransaction(format!(
-                        "Failed to deserialize lookup table: {}",
-                        sanitize_error!(e)
-                    ))
-                })?;
+        for lookup in lookup_table_lookups {
+            full_address_lists.push(
```
dev-jodee: Now that the branches are unified you can drop full_address_lists entirely: all the cache inserts happen above, so the loop below can just do let addresses = cache.get(&lookup.account_key).ok_or(...)? and index into the borrow. Saves a full Vec clone per lookup.

dev-jodee (changes requested):

raushan728:
Review address form 13646d7bc8c972680a5b44c9366a6b6b2296d8df

dev-jodee (approved):

jacobcreech (approved):

dev-jodee:
sorry @raushan728 github rebased the pr but un-signed all the commits, mind re-signing them

raushan728:
> sorry @raushan728 github rebased the pr but un-signed all the commits, mind re-signing them

I'll do it tomorrow as I've already shut down my system and am heading to sleep now. 🙂

dev-jodee (changes requested):
sign

raushan728:
> sign

Done!

dev-jodee (approved):

dev-jodee:
thanks @raushan728

PR #571, opened 2026-06-19
fix(usage-limit): make multi-rule limit enforcement all-or-nothing

raushan728:
Sequential `check_and_increment` calls per rule left earlier rules incremented when a later rule denied denied requests could silently consume quota under concurrent load.

* Added `check_and_increment_many` to `UsageStore` trait with atomic Redis (Lua) and in-memory (single mutex) backends; default impl keeps external implementations from breaking
* `check_and_record` deduplicates entries by key and issues one batch call instead of a sequential loop
* Per-rule denial reason restored on the batch-denial path
* Added in-memory and Redis integration tests (`#[ignore]`) for both Lua scripts

Builds on #463.

### DIFF
--- a/crates/lib/src/usage_limit/usage_store.rs
+++ b/crates/lib/src/usage_limit/usage_store.rs
@@ -29,6 +29,12 @@ pub trait UsageStore: Send + Sync {
         expiry: Option<u64>,
     ) -> Result<bool, KoraError>;
 
+    /// `entries` must contain distinct keys; duplicate keys produce undefined increment behaviour.
+    async fn check_and_increment_many(
+        &self,
+        entries: &[(String, u64, u64, Option<u64>)],
+    ) -> Result<bool, KoraError>;
+
     /// Clear all usage data (mainly for testing)
     async fn clear(&self) -> Result<(), KoraError>;
 }
@@ -44,6 +50,31 @@ static CHECK_AND_INCREMENT_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
             redis.call('INCRBY', KEYS[1], ARGV[1])
             if ARGV[3] ~= '0' and redis.call('TTL', KEYS[1]) < 0 then
                 redis.call('EXPIREAT', KEYS[1], ARGV[3])
+            end
+            return 1
+            ",
+    )
+});
+
+/// Atomically checks limits for multiple rules and increments only if all pass.
+static CHECK_AND_INCREMENT_MANY_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
+    redis::Script::new(
+        r"
+            local num_keys = #KEYS
+            for i = 1, num_keys do
+                local current = redis.call('GET', KEYS[i])
+                local count = current and tonumber(current) or 0
+                local delta = tonumber(ARGV[(i - 1) * 3 + 1])
+                local max = tonumber(ARGV[(i - 1) * 3 + 2])
+                if count + delta > max then return 0 end
+            end
+            for i = 1, num_keys do
+                local delta = tonumber(ARGV[(i - 1) * 3 + 1])
+                local expiry = ARGV[(i - 1) * 3 + 3]
+                redis.call('INCRBY', KEYS[i], delta)
+                if expiry ~= '0' and redis.call('TTL', KEYS[i]) < 0 then
+                    redis.call('EXPIREAT', KEYS[i], expiry)
+                end
             end
             return 1
             ",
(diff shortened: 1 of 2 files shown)


crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -314,10 +313,13 @@ impl UsageTracker {
...
+            let entries: Vec<_> =
+                pending_increments.iter().map(|(k, d, m, e, _)| (k.clone(), *d, *m, *e)).collect();
+
+            if !self.store.check_and_increment_many(&entries).await? {
                 return Ok(LimiterResult::Denied {
-                    reason: format!("User {} exceeded {} limit", ctx.user_id, description),
+                    reason: format!("User {} exceeded usage limit", ctx.user_id),
                 });
             }
         }
```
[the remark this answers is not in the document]
raushan728: In the batch path, the failure is atomic there's no deterministic "which rule fired" since all checks happen together before any increment. The fast-path pre-check above still emits detailed per-rule messages for the common case. Generic message here is intentional.

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -314,10 +313,13 @@ impl UsageTracker {
             pending_increments.push((key, increment_count, max, expiry, description));
         }
 
-        for (key, increment_count, max, expiry, description) in pending_increments {
-            if !self.store.check_and_increment(&key, increment_count, max, expiry).await? {
+        if !pending_increments.is_empty() {
+            let entries: Vec<_> =
```
dev-jodee, line 330: Nothing guarantees distinct keys here (dup rules in config, or two windowed rules on the same bucket), so with dup keys both checks read the same count and both increments apply, so the limit can be exceeded. The old sequential loop handled this. Think we need to merge entries by key?

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -314,10 +313,13 @@ impl UsageTracker {
...
+                pending_increments.iter().map(|(k, d, m, e, _)| (k.clone(), *d, *m, *e)).collect();
+
+            if !self.store.check_and_increment_many(&entries).await? {
                 return Ok(LimiterResult::Denied {
-                    reason: format!("User {} exceeded {} limit", ctx.user_id, description),
+                    reason: format!("User {} exceeded usage limit", ctx.user_id),
```
dev-jodee, line 349: Now the denial reason is all behind the same generic message, will be hard to debug why someone was denied

crates/lib/src/usage_limit/usage_store.rs
```diff
@@ -50,6 +56,31 @@ static CHECK_AND_INCREMENT_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
     )
 });
 
+/// Atomically checks limits for multiple rules and increments only if all pass.
+static CHECK_AND_INCREMENT_MANY_SCRIPT: Lazy<redis::Script> = Lazy::new(|| {
```
dev-jodee, line 68: wondering since the LUA script is getting more and more complex, I think an integration test with actual redis (like a quick redis service in the ci) to have integration test, rather than only for the in memory, would be great

crates/lib/src/usage_limit/usage_store.rs
```diff
@@ -29,6 +29,12 @@ pub trait UsageStore: Send + Sync {
         expiry: Option<u64>,
     ) -> Result<bool, KoraError>;
 
+    /// `entries` must contain distinct keys; duplicate keys produce undefined increment behaviour.
+    async fn check_and_increment_many(
```
dev-jodee, line 34: adding required func will break all current external implementations (i.e. from other teams that have custom impl). We should have a default impl for this, maybe that loops over the values and does an increment on them. Won't be atomic but at least it wont break the external impls

dev-jodee (changes requested):

crates/lib/src/usage_limit/usage_store.rs
raushan728: I added radis integration tests with `#[ignore]`, and for CI, a separate PR would be cleaner - I haven't looked at the workflow yet but happy to open it. We can hold this until that lands if you'd prefer

crates/lib/src/usage_limit/usage_tracker.rs
```diff
@@ -314,10 +313,40 @@ impl UsageTracker {
...
+                if let Some(existing) = unique_pending.iter_mut().find(
+                    |(k, _, _, _, _): &&mut (String, u64, u64, Option<u64>, String)| *k == key,
+                ) {
+                    existing.1 += delta;
+                    existing.2 = min(existing.2, max);
+                } else {
+                    unique_pending.push((key, delta, max, expiry, desc));
+                }
+            }
+            let pending_increments = unique_pending;
```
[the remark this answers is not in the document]
raushan728: The `min(max)` dedup and sequential default are both intentional

crates/lib/src/usage_limit/usage_store.rs
raushan728: @dev-jodee I looked into the CI setup `rust-unit.yml` runs unit tests there are no service containers configured yet. Plan is to add a redis service container and run integration tests with `--include-ignored test_redis` to target only redis tests by name. One consern there's already an existing `#[ignore]` test in `discovery.rs` that makes a live devnet RPC call and more could be added in future. So using `--include-ignored` globally seems risky. Targeting by name feels safer but wanted your input before opening the PR.

dev-jodee (approved):

PR #578, opened 2026-06-25
feat(validation): support CreateAccountAllowPrefund instruction

dev-jodee:
## Summary
- The System program's `CreateAccountAllowPrefund` was absent from `solana-system-interface 2.0.0`, so it fell through the instruction parser's catch-all and silently bypassed the fee-payer policy and outflow accounting. The parser now decodes its bincode wire layout directly (variant tag 13) and maps it onto the existing `SystemCreateAccount` parsed type, so validation and fee logic apply unchanged.
- Added `system_create_account_allow_prefund` index module — its account order is the reverse of `CreateAccount` (`[new, funder]`) and the funder is omitted when `lamports == 0`.
- Because a prefunded target can already hold lamports, the fee payer can be the account being created (and bricked by allocate+assign) without being the funder the existing gate checks. That vector is now rejected under the same `allow_create_account` policy.
- A stack bump to the upstream enum was evaluated and rejected: `system-interface 3.x` requires `solana-address 2.x` / `pubkey 4.x`, but `solana-loader-v4-interface` has no pubkey-4 release, so no consistent stable dependency set exists yet.

## Test Plan
- `cargo test -p kora-lib --lib` — 829 pass, incl. 5 new tests.
- Tag anchor test serializes the real `UpgradeNonceAccount` (= 12) to prove the appended `CreateAccountAllowPrefund` tag is 13; fails loudly if upstream reorders the enum.
- Parse tests cover the 2-account (funder) and 1-account (`lamports == 0`) forms.
- Validator tests cover the funder vector, the brick vector, and the no-funding brick vector across both policy states.

### DIFF
--- a/crates/lib/src/transaction/instruction_util.rs
+++ b/crates/lib/src/transaction/instruction_util.rs
@@ -2009,12 +2009,12 @@ impl IxUtils {
         Ok(parsed_instructions)
     }
 
-    // Absent from system-interface 2.0.0; decode tag 13 (tag, lamports, space, owner) by hand.
+    // Absent from system-interface 2.0.0; decode (tag, lamports, space, owner) by hand.
     fn parse_create_account_allow_prefund(data: &[u8]) -> Option<(u64, Pubkey)> {
-        const CREATE_ACCOUNT_ALLOW_PREFUND_TAG: u32 = 13;
         let (tag, lamports, _space, owner) =
             bincode::deserialize::<(u32, u64, u64, Pubkey)>(data).ok()?;
-        (tag == CREATE_ACCOUNT_ALLOW_PREFUND_TAG).then_some((lamports, owner))
+        (tag == instruction_indexes::system_create_account_allow_prefund::DISCRIMINATOR)
+            .then_some((lamports, owner))
     }
 
     pub fn parse_alt_instructions(
@@ -3430,7 +3430,10 @@ impl IxUtils {
 mod tests {
 
     use super::*;
-    use crate::transaction::versioned_transaction::VersionedTransactionResolved;
+    use crate::{
+        constant::instruction_indexes::system_create_account_allow_prefund::DISCRIMINATOR,
+        transaction::versioned_transaction::VersionedTransactionResolved,
+    };
     use solana_sdk::{
         instruction::{AccountMeta, Instruction},
         message::{AccountKeys, Message, VersionedMessage},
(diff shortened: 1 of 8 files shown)


crates/lib/src/validator/transaction_validator.rs
```diff
@@ -2986,6 +3005,78 @@ mod tests {
...
+        let build_ix =
+            |new_account: Pubkey, funder: Pubkey, lamports: u64, owner: Pubkey| -> Instruction {
+                let mut accounts = vec![AccountMeta::new(new_account, true)];
+                if lamports > 0 {
+                    accounts.push(AccountMeta::new(funder, true));
+                }
+                Instruction {
+                    program_id: SYSTEM_PROGRAM_ID,
+                    accounts,
+                    data: bincode::serialize(&(13u32, lamports, 100u64, owner)).unwrap(),
```
amilz: assuming we need to do this b/c it's not in sdk yet. fine, butbut might nice to have discriminator as a const

crates/lib/src/validator/transaction_validator.rs
```diff
@@ -316,6 +316,25 @@ impl TransactionValidator {
             }
         });
 
+        // Prefund lets the fee payer be the account created (bricked), not just the funder above.
+        // Re-run the same gate keyed on new_account.
+        validate_system!(self, system_instructions, SystemCreateAccount,
+        ParsedSystemInstructionData::SystemCreateAccount { new_account, owner, .. } => new_account,
```
amilz, line 325: don't we already have a block this can fit in? 
just confirming that `SystemCreateAccount` here wont' miss our intended new ix?

crates/lib/src/transaction/instruction_util.rs
amilz, line 1: just confirming that cpi to the new ix works okay? if `reconstruct_system_instruction` doesn't have a matching `createAccountAllowPrefund`, do we throw?

amilz (dismissed):
lgtm. could of nit comments below.

also just want to confirm there's nothing we need to do here wrt `feepayer_outflow` for `SystemAssign`/`SystemAllocate`? just trying to think how someone might user this w/ kora nad what implciations might be for fee calculation

crates/lib/src/validator/transaction_validator.rs
dev-jodee: yeah, and if we upgrade those packages....it all implodes

crates/lib/src/transaction/instruction_util.rs
dev-jodee: Yeah wont be missed, but I added a test to confirm it wont be !

amilz (dismissed):

amilz (approved):

PR #579, opened 2026-06-27
feat(pricing): add FixedStable price model for multi-stablecoin fixed fees

amilz:
Operators can now specify a single fixed fee that applies to any of a
list of stablecoin mints, treating them as equivalent (e.g. charge
0.10 regardless of whether the user pays in USDC, USDT, PYUSD, etc.).

- Add `PriceModel::FixedStable { amount, tokens, strict }` variant
- Fee quoting returns `amount` directly for any token in the group,
  bypassing per-token oracle calls
- Lamport equivalent still uses the first listed token as reference
  (one oracle call for validation/max_allowed_lamports enforcement)
- Add `get_required_lamports_with_fixed_stable` and
  `get_fixed_stable_fee_for_token` to PriceConfig
- Handle FixedStable in estimate_kora_fee and calculate_fee_in_token
- Update OpenAPI spec with fixed_stable variant
- Add TOML builder helper and full unit test coverage

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -818,6 +818,27 @@ impl Config {
             ))
         })?;
 
+        // Validate FixedStable price model configuration
+        if let PriceModel::FixedStable { amount, tokens, .. } = &config.validation.price.model {
+            if tokens.is_empty() {
+                return Err(KoraError::ConfigError(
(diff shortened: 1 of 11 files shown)


amilz:
not needed. can you use `fixed` and specify different acceptable payment tokens in config

PR #590, opened 2026-07-06
fix(transaction): allow Token-2022 metadata and group instructions

### LINKED ISSUE #588: Token-2022 parsing rejects token-metadata interface instructions (TokenMetadataInitialize) with 0xc InvalidInstruction
## Summary

Kora rejects any sponsored transaction containing a **token-metadata interface instruction** (e.g. `TokenMetadataInitialize`, which writes a Token-2022 mint's name/symbol/uri) with:

```
-32000 Invalid transaction: Failed to parse Token-2022 instruction: Custom program error: 0xc
```

`0xc` = `TokenError::InvalidInstruction` (12). The transaction is valid — it lands fine when self-submitted without Kora. This blocks sponsoring essentially any Token-2022 issuance flow that sets on-chain metadata.

## Reproduction

Token-2022 token deploys built with [`@solana/mosaic-sdk`](https://www.npmjs.com/package/@solana/mosaic-sdk), fee-sponsored through Kora. The deploy's metadata step emits a `TokenMetadataInitialize` instruction addressed to the Token-2022 program; every `signAndSendTransaction` for such a deploy returns the `0xc` error above (reproduced on `v2.2.0-beta.7`).

## Root cause

The Token-2022 program processes **two distinct instruction encodings**:

1. **`TokenInstruction`** (`spl-token-2022-interface`) — the core enum, 1-byte discriminants.
2. **Interface instructions** (`spl-token-metadata-interface`, token-group, …) — separate crates using 8-byte "splat" discriminators (e.g. `Initialize::SPL_DISCRIMINATOR_SLICE`), routed to the *same* program id.

Kora's parser assumes vocabulary 1 is exhaustive (`crates/lib/src/transaction/instruction_util.rs`):

```rust
} else if program_id == spl_token_2022_interface::ID {
    let spl_ix = spl_token_2022_interface::instruction::TokenInstruction::unpack(
        &instruction.data,
    )
    .map_err(|e| { ... })?;
    // ← TokenMetadataInitialize's 8-byte discriminator reads as garbage → 0xc → abort
```

The same assumption exists in `Token2022SecurityParser::parse` (`token2022_security.rs`).

Kora already knows the TokenMetadata *extension type* (for account-size math) — just not the *instruction family* that populates it.

## Fix

At both sites, when `TokenInstruction::unpack` fails, try `spl_token_metadata_interface::instruction::TokenMetadataInstruction::unpack`:

- Main parser: route valid metadata-interface instructions through the existing `SplTokenUnknownExtension` handling so every touched account is still policy-validated.
- Security parser: skip them (they carry none of the security-relevant extension initializations that parser models).

Anything that parses as *neither* vocabulary still hard-errors — the fail-closed posture is unchanged. PR incoming.

multipletwigs:
Fixes #588

## Summary

Kora was rejecting valid Token-2022 transactions that include token metadata or token group instructions.

These instructions are sent to the Token-2022 program, but they are not normal `TokenInstruction` variants. Because Kora only tried `TokenInstruction::unpack`, valid instructions like `TokenMetadataInitialize` failed with:

```text
-32000 Invalid transaction: Failed to parse Token-2022 instruction: Custom program error: 0xc
```

This PR lets Kora recognize those Token-2022 interface instructions and still run the same fee-payer safety checks around touched accounts and authority fields.

## What Changed

- Recognize `spl-token-metadata-interface` instructions routed to Token-2022.
- Recognize `spl-token-group-interface` instructions routed to Token-2022.
- Skip those instructions in the main typed SPL parser after confirming they are valid interface instructions.
- Validate their accounts and authority fields in `Token2022SecurityParser`.
- Keep rejecting unknown Token-2022 instruction bytes.
- Add direct workspace dependencies for `spl-token-metadata-interface` and `spl-token-group-interface`


multipletwigs:
Addressed the token-group gap in 7e78cbe: the check is now a shared `IxUtils::is_token2022_interface_instruction` helper (used by both parser sites, so they can't drift) that also recognizes `spl-token-group-interface` instructions. Added tests for the group instructions at both sites — each verified to fail without the guard — plus fail-closed tests asserting unrecognized Token-2022 data is still rejected.

crates/lib/src/transaction/token2022_security.rs
```diff
@@ -74,12 +74,23 @@ impl Token2022SecurityParser {
...
-            })?;
+            let token_instruction = match TokenInstruction::unpack(&instruction.data) {
+                Ok(token_instruction) => token_instruction,
+                Err(e) => {
+                    // Interface instructions (spl-token-metadata-interface and
+                    // spl-token-group-interface) are not TokenInstruction variants and
+                    // carry none of the security-relevant extension initializations this
+                    // parser looks for; their accounts are validated by the main
+                    // instruction parser's unknown-extension path.
+                    if IxUtils::is_token2022_interface_instruction(&instruction.data) {
```
dev-jodee: I think this opens up potential attack vectors to use the kora auth as the current authority in the instruction right ? (like UpdateField, RemoveKey, etc.?)

dev-jodee (commented):
wondering if this makes it so some auth ix related to metadata & group are now bypassed?

crates/lib/src/transaction/token2022_security.rs
multipletwigs: Hm yeah gonna rework it a bit

crates/lib/src/transaction/token2022_security.rs
```diff
@@ -169,6 +185,111 @@ impl Token2022SecurityParser {
...
+                .collect(),
+            ),
+            TokenMetadataInstruction::Emit(_) => ("Token2022 TokenMetadataEmit", vec![]),
+        };
+
+        Token2022SecurityInstruction {
+            instruction_name,
+            extension_type: Some(ExtensionType::TokenMetadata),
+            accounts: Self::instruction_accounts(instruction),
+            account_usage_policy: Token2022AccountUsagePolicy::RejectIfFeePayerPresent,
```
dev-jodee: So for this policy, I think it's a bit too strict, since we have a config 

`fee_payer_policy.token_2022.allow_update_extension_authority`

if that's set as enabled, we should allow the instruction right? It's the same as updating the extension (update the metadata within it), same for the group pointer a bit lower, wdyt ?

Image

dev-jodee (changes requested):
looks better, but now a bit too strict :P

multipletwigs:
@dev-jodee addressed the issues and now it should properly follow the configs :) thank you for helping out

crates/lib/src/validator/transaction_validator.rs
dev-jodee, line 1: extension_type blocking branch unreachable — no config knob to deny metadata/group. Can never evaluate true for the three types #590 uses. Before the pr, these were blanket-rejected; post-PR there's no way to deny them. Inconsistent with #628's convention in the same file. (line 853)

dev-jodee (changes requested):
one small thing ot harden

crates/lib/src/validator/transaction_validator.rs
multipletwigs: taking a look!

multipletwigs: does it make sense to have token_metadata, token_group and token_group_member be first class citizens of blocked_mint_extensions config? We extend the arm so operators can now treat them the same as metadata_pointer. Makes it sponsored by default, but deniable via config.

multipletwigs:
@dev-jodee requesting another run on ci ! also the @greptile-apps everything is already properly signed

dev-jodee (approved):

amilz (approved):

PR #591, opened 2026-07-07
feat(validation): support p-token batch and lamport instructions

dev-jodee:
## Summary
- p-token activates at the SPL Token program ID and adds three instructions — `Batch` (255), `WithdrawExcessLamports` (38), `UnwrapLamports` (45). The SPL instruction parser hit a `_ => {}` catch-all and silently dropped them.
- **Security:** a `Batch` could wrap any fee-payer-restricted instruction (e.g. a Transfer with the fee payer as owner) and bypass `fee_payer_policy` entirely. `Batch` sub-instructions are now decoded via the exact on-chain wire format and run through the existing parser, so each is validated individually. Nested and malformed batches are rejected (fail-closed), matching the program's own rule.
- Added `fee_payer_policy` controls `allow_withdraw_excess_lamports` and `allow_unwrap_lamports` on both `spl_token` and `token_2022`. Field-level `#[serde(default)]` so existing configs parse unchanged; default `false` (restrictive), matching the existing derived-`Default` convention. Only triggers when the fee payer is the instruction authority.
- No new token program or `TokenType` variant — p-token is the same program ID, so existing transfer / ATA / fee-estimation logic works unchanged.

## Test Plan
- `cargo test -p kora-lib` — 837 unit tests pass, including:
  - `test_fee_payer_policy_spl_transfer_wrapped_in_batch_is_enforced` — regression guard: a batch-wrapped transfer by the fee payer must still be rejected under `allow_transfer=false`, and allowed when the policy permits it.
  - parse tests for batch extraction, nested-batch rejection, and `WithdrawExcessLamports` / `UnwrapLamports`.
- `cargo fmt --check` and `cargo clippy` clean.
- Integration tests (`just integration-test`) not run here — needs a local validator.

## Notes
- Pre-existing, out of scope: `sdks/ts/src/client.ts:407` builds the payment transfer without passing `programAddress`, so it hardcodes SPL-Token even for Token-2022 fee tokens. Unrelated to p-token (the payment transfer already works for p-token, same program ID).


crates/lib/src/tests/config_mock.rs
```diff
@@ -716,6 +718,7 @@ impl FeePayerPolicyBuilder {
                     allow_initialize_mint: false,
                     allow_initialize_account: false,
                     allow_initialize_multisig: false,
+                    allow_withdraw_excess_lamports: false,
```
amilz, line 721:
```suggestion
                    allow_withdraw_excess_lamports: false,
                    allow_unwrap_lamports: false
```

crates/lib/src/validator/config_validator.rs
```diff
@@ -2435,6 +2437,7 @@ mod tests {
                         allow_update_extension_authority: true,
                         allow_freeze_account: true,
                         allow_thaw_account: true,
+                        allow_withdraw_excess_lamports: true,
```
amilz, line 2440: I think just add across the board, no?

```suggestion
                        allow_withdraw_excess_lamports: true,
                         allow_unwrap_lamports: true,
```

crates/lib/src/validator/transaction_validator.rs
```diff
@@ -420,6 +420,18 @@ impl TransactionValidator {
...
+            validate_spl!(self, spl_instructions, SplTokenWithdrawExcessLamports,
+                ParsedSPLInstructionData::SplTokenWithdrawExcessLamports { owner, multisig_signers, is_2022 } => { owner, multisig_signers, is_2022 },
+                self.fee_payer_policy.spl_token.allow_withdraw_excess_lamports,
+                self.fee_payer_policy.token_2022.allow_withdraw_excess_lamports,
+                "SPL Token WithdrawExcessLamports", "Token2022 Token WithdrawExcessLamports");
+
+            validate_spl!(self, spl_instructions, SplTokenUnwrapLamports,
+                ParsedSPLInstructionData::SplTokenUnwrapLamports { owner, multisig_signers, is_2022 } => { owner, multisig_signers, is_2022 },
+                self.fee_payer_policy.spl_token.allow_unwrap_lamports,
+                self.fee_payer_policy.spl_token.allow_unwrap_lamports,
```
amilz: should one of these be t22?

crates/lib/src/config.rs
```diff
@@ -306,6 +306,12 @@ pub struct SplTokenInstructionPolicy {
     pub allow_freeze_account: bool,
     /// Allow fee payer to be the freeze authority in SPL Token ThawAccount instructions
     pub allow_thaw_account: bool,
+    /// Allow fee payer to be the authority in SPL Token WithdrawExcessLamports instructions
+    #[serde(default)]
+    pub allow_withdraw_excess_lamports: bool,
+    /// Allow fee payer to be the authority in SPL Token UnwrapLamports instructions
+    #[serde(default)]
+    pub allow_unwrap_lamports: bool,
```
amilz, line 314: could it be useful to disallow batches all together?

tests/fee_payer_policy/fee_payer_policy_violations.rs
```diff
@@ -426,6 +426,130 @@ async fn test_revoke_token2022_policy_violation() {
...
+            error.assert_contains_message(
+                "Fee payer cannot be used for 'Token2022 Token WithdrawExcessLamports'",
+            );
+        }
+        Ok(_) => panic!("Expected error for Token2022 WithdrawExcessLamports policy violation"),
+    }
+}
+
+#[tokio::test]
+async fn test_batch_wrapped_transfer_does_not_bypass_policy() {
```
amilz, line 509: wonder if it'd make sense to do a CPI test w/ a batch as well.

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -831,6 +848,83 @@ impl IxUtils {
...
+    fn is_spl_token_batch(instruction: &Instruction) -> bool {
+        instruction.program_id == spl_token_interface::ID
+            && instruction.data.first() == Some(&BATCH_DISCRIMINATOR)
+    }
+
+    fn expand_spl_token_batches(
+        instructions: &[Instruction],
+    ) -> Result<Vec<Instruction>, KoraError> {
+        let mut expanded = Vec::with_capacity(instructions.len());
+        for instruction in instructions {
```
amilz, line 874: i feel like since batch is the going to be called less often than most, you might do some check up front to avoid cloning over each ix. something like...

```rs
    if !instructions.iter().any(Self::is_spl_token_batch) {
        return Ok(instructions);
    }
   // existing let mut expanded, etc...
```

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -831,6 +848,83 @@ impl IxUtils {
         );
     }
 
+    fn is_spl_token_batch(instruction: &Instruction) -> bool {
+        instruction.program_id == spl_token_interface::ID
+            && instruction.data.first() == Some(&BATCH_DISCRIMINATOR)
+    }
```
amilz, line 855: I guess batch isn't on t22? If not, should we throw here if if program id = t22? just so batches don't go unchecked once batch exists in t22? then we can update here?

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -831,6 +848,83 @@ impl IxUtils {
...
+            let sub_accounts =
+                batch.accounts[account_cursor..account_cursor + account_count].to_vec();
+            account_cursor += account_count;
+
+            if sub_data.first() == Some(&BATCH_DISCRIMINATOR) {
+                return Err(KoraError::InvalidTransaction(
+                    "Nested p-token batch instructions are not allowed".to_string(),
+                ));
+            }
+
```
amilz, line 924: i suspect token program wouild catch a bad ix, but we could just do a check here? 

```rs
            if spl_token_interface::instruction::TokenInstruction::unpack(&sub_data).is_err() {
                return Err(KoraError::InvalidTransaction(
                    "Malformed p-token batch: unrecognized sub-instruction".to_string(),
                ));
            }
```

crates/lib/src/transaction/instruction_util.rs
```diff
@@ -831,6 +848,83 @@ impl IxUtils {
...
+            if data_cursor + data_len > data.len() {
+                return Err(KoraError::InvalidTransaction(
+                    "Malformed p-token batch: sub-instruction data out of bounds".to_string(),
+                ));
+            }
+            let sub_data = data[data_cursor..data_cursor + data_len].to_vec();
+            data_cursor += data_len;
+
+            if account_cursor + account_count > batch.accounts.len() {
```
amilz, line 910: does batch have fail cases we should check for here to prevent sending bad txns (e.g., batch len > 0, no data)

amilz (commented):
flagged a few spot where upwrap should be in t22, i think. didn't comment after the 1st few.

crates/lib/src/config.rs
dev-jodee: I think it could be something we add if we see demand, but for now I can't think of a use case, since we already validate all "sub" instructions of the batch

crates/lib/src/transaction/instruction_util.rs
dev-jodee: Cases were fixed from the other comment and len < 0 will just skip this

amilz (dismissed):

amilz (approved):

amilz (approved):
p-token batch/lamport validation with fail-closed batch decoding and new fee_payer_policy controls; all CI green, 837 unit tests pass incl. batch-wrapped-transfer regression guard.

PR #592, opened 2026-07-08
feat(kora-deploy): add registry-gated program upgrades

dev-jodee:
## Summary
- **kora-deploy can now upgrade programs, not just deploy them** — reruns in the same directory auto-detect deploy vs upgrade via a `kora-deploy.json` manifest + on-chain existence check.
- **Closes an existing authorization hole**: previously *any* client could upgrade or close *any* program deployed through the paymaster, because the `deploy_authority` plugin only enforced that Kora stayed the loader authority, never who the deployer was.
- **On-chain deploy registry** (`examples/devnet-deploy-paymaster/registry-program/`) binds each program to its deployer's wallet in a PDA keyed by program id. Registration is atomic in the deploy tx and gated by the program keypair's signature, so ownership can't be forged or front-run.
- **Plugin gating**: upgrade, close, and loader-v4 mutations on an *existing* program now require a registry entry **and** the registered owner's verified signature. Unregistered programs are immutable through the paymaster (backward-compatible: deploying without `--wallet` still works, just non-upgradeable).
- **Reaper** reclaims registry-entry rent when it closes idle programs.

## Test Plan
- `cargo test -p kora-lib` — 845 pass, incl. 10 new registry-gating tests (unregistered reject, wrong-wallet reject, owner-sig accept, close gating, config validation).
- Full local e2e (validator + Kora with plugin live): deploy → upgrade → close all succeed as registered owner; **0/16 adversarial probes signed**, including three new hijack attempts (upgrade w/o owner sig, upgrade by wrong wallet, close w/o owner sig) — all rejected.
- `cargo clippy` clean on `kora-deploy` and the example crate.

## Notes
- Pre-existing clippy findings in `crates/lib/src/validator/config_validator.rs` (`std::slice::from_ref` lint) are unrelated to this change and left untouched.
- The reaper's registry-entry cleanup is compile-checked and logically sound but not exercised end-to-end (needs a 7-day-idle program to trigger).


crates/kora-deploy/src/main.rs
```diff
@@ -1,14 +1,18 @@
...
+    pubkey::Pubkey,
+    signature::{read_keypair_file, Keypair},
+    signer::Signer,
+};
 
 #[derive(Parser)]
 #[command(
     name = "kora-deploy",
-    about = "Deploy a Solana program to devnet via a Kora paymaster (no SOL required)."
+    about = "Deploy or upgrade a Solana program on devnet via a Kora paymaster (no SOL required)."
```
amilz, line 15: personal pref would be to have `kora-deploy` and `kora-upgrade` or `kora-deploy upgrade` when upgrading

crates/kora-deploy/src/main.rs
```diff
@@ -24,36 +28,117 @@ struct Args {
...
+        args.user_id.clone().unwrap_or_else(|| format!("kora-deploy-{}", Pubkey::new_unique()));
+    let wallet = load_wallet(args.wallet.as_deref())?;
+
+    match args.program_id {
+        Some(program) => {
+            if !program_is_live(&args.rpc_url, &program).await? {
+                bail!(
+                    "program {program} is not live on-chain (never deployed or reaped); \
+                     omit --program-id to deploy fresh"
+                );
```
amilz, line 56: see comment above. idk feels like a clunky ux.

amilz (dismissed):

crates/kora-deploy/src/main.rs
dev-jodee: Think it's easier for automated tool / ci to use one command, as they are not chain aware necessarily. For now will leave like this and see if there's feedback on any preference

amilz (approved):

PR #601, opened 2026-07-09
chore(kora-deploy): bump to 0.2.0 for crates.io publish

dev-jodee:
## Summary
- Bump `kora-deploy` `0.1.0 → 0.2.0`. `0.1.0` (the deploy-only version) is already on the crates.io index, so the registry-gated upgrade support (#592) can't publish under it.
- Updates the workspace dependency pin and `Cargo.lock` to match.

## Why
The publish workflow failed with `crate kora-deploy@0.1.0 already exists on crates.io index`. This bump unblocks publishing the upgrade-capable client.

## Test Plan
- `cargo check -p kora-deploy -p devnet-deploy-paymaster` passes with the new version.
- After merge: re-run the `rust-publish` workflow with `publish-kora-deploy=true` to publish 0.2.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3547,7 +3547,7 @@ dependencies = [
 
 [[package]]
 name = "kora-deploy"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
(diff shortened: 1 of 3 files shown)


amilz (approved):
kora-deploy 0.1.0→0.2.0 version bump to unblock crates.io publish; all CI green. LGTM.

PR #603, opened 2026-07-10
fix: align fee estimation and config validation with runtime behavior

dev-jodee:
## Summary
- **Fee estimation**: recognize a payment whose destination ATA is created in the same transaction. The estimator only read pre-state accounts, so an in-transaction Token-2022 payment ATA read as no-payment — it skipped the transfer-fee surcharge and diverged from payment validation. Now falls back to `find_ata_creation_for_destination` on `AccountNotFound`, matching validation.
- **Config validation**: reflect env-overridden auth secrets. Runtime resolves auth env-first over `kora.toml`, but `config validate` only inspected the TOML, so it could report the config secret while a stale env var was active, or warn "no auth configured" while auth was live. Adds a shared `AuthConfig::resolve_secret` used by both server and validation, and warns when an env var overrides a differing `kora.toml` value (never logs secret contents). Env precedence unchanged.
- **Usage limits**: refresh the window timestamp at store time. The window bucket key/expiry were derived from a timestamp captured before validation+RPC, so under latency the bucket could be written already-expired and the windowed limit would not be enforced. Fix stamps the current time just before the store phase. (This also fixes the intermittent `Integration Tests (usage_limit)` flake on windowed-limit tests.)
- **Docs**: note that `user_id` is trusted app input for usage limits.

## Test Plan
- `cargo test -p kora-lib fee::` — 52 passed (incl. in-transaction payment ATA test)
- `cargo test -p kora-lib config_validator` — 66 passed (incl. env-override warning test)
- `cargo test -p kora-lib usage_limit` — 57 passed (incl. `test_windowed_limit_not_bypassed_by_stale_timestamp`)

### DIFF
--- a/crates/lib/src/fee/fee.rs
+++ b/crates/lib/src/fee/fee.rs
@@ -11,7 +11,6 @@ use crate::{
     token::{
         spl_token_2022::Token2022Mint,
         token::{AtaCreationInstructionInfo, TokenType, TokenUtil, TransferHookValidationFlow},
-        TokenState,
     },
     transaction::{
         ParsedALTInstructionData, ParsedALTInstructionType, ParsedSPLInstructionData,
@@ -97,36 +96,6 @@ impl FeeConfigUtil {
 
     /// Helper function to check if a token transfer instruction is a payment to Kora
     /// Returns Some(token_account_data) if it's a payment, None otherwise
-    async fn get_payment_instruction_info(
-        config: &Config,
-        rpc_client: &RpcClient,
-        destination_address: &Pubkey,
-        payment_destination: &Pubkey,
-        skip_missing_accounts: bool,
-    ) -> Result<Option<Box<dyn TokenState + Send + Sync>>, KoraError> {
-        // Get destination account - handle missing accounts based on skip_missing_accounts
-        let destination_account =
-            match CacheUtil::get_account(config, rpc_client, destination_address, false).await {
-                Ok(account) => account,
-                Err(_) if skip_missing_accounts => {
-                    return Ok(None);
-                }
-                Err(e) => {
-                    return Err(e);
-                }
-            };
-
-        let token_program = TokenType::get_token_program_from_owner(&destination_account.owner)?;
-        let token_account = token_program.unpack_token_account(&destination_account.data)?;
-
-        // Check if this is a payment to Kora
-        if token_account.owner() == *payment_destination {
-            Ok(Some(token_account))
-        } else {
-            Ok(None)
-        }
-    }
-
     /// Analyze payment instructions in transaction
     /// Returns (has_payment, total_transfer_fees)
     async fn analyze_payment_instructions(
@@ -139,6 +108,7 @@ impl FeeConfigUtil {
         let mut has_payment = false;
         let mut total_transfer_fees = 0u64;
 
+        let all_instructions = resolved_transaction.all_instructions.clone();
         let parsed_spl_instructions = resolved_transaction.get_or_parse_spl_instructions()?;
 
         for instruction in parsed_spl_instructions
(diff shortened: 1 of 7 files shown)


crates/lib/src/fee/fee.rs
```diff
@@ -97,36 +96,6 @@ impl FeeConfigUtil {
...
-        if token_account.owner() == *payment_destination {
-            Ok(Some(token_account))
-        } else {
-            Ok(None)
-        }
-    }
-
     /// Analyze payment instructions in transaction
     /// Returns (has_payment, total_transfer_fees)
     async fn analyze_payment_instructions(
```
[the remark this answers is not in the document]
dev-jodee: Fixed in f49594834 — removed the two leftover lines from the deleted `get_payment_instruction_info` helper. Kept the existing plain-doc style rather than the backtick reformat.

crates/lib/src/validator/config_validator.rs
```diff
@@ -1222,6 +1232,93 @@ mod tests {
...
+            metrics: MetricsConfig::default(),
+        };
+        let _ = update_config(config);
+
+        let rpc_client = RpcClient::new_with_commitment(
+            "http://localhost:8899".to_string(),
+            CommitmentConfig::confirmed(),
+        );
+        let result = ConfigValidator::validate_with_result(&rpc_client, true).await;
+        std::env::remove_var(AuthConfig::API_KEY_ENV);
```
amilz, line 1283: do you need to remove Jup key?

crates/lib/src/fee/fee.rs
```diff
@@ -139,6 +106,7 @@ impl FeeConfigUtil {
         let mut has_payment = false;
         let mut total_transfer_fees = 0u64;
 
+        let all_instructions = resolved_transaction.all_instructions.clone();
```
amilz, line 111: do we need to include bundles here?
 
```suggestion
let all_instructions = bundle_instructions
    .map(|ix| ix.to_vec())
    .unwrap_or_else(|| resolved_transaction.all_instructions.clone());
```

crates/lib/src/fee/fee.rs
```diff
@@ -153,17 +121,30 @@ impl FeeConfigUtil {
...
-                    &payment_destination,
-                    true, // Skip missing accounts
-                )
-                .await?;
+                // Resolve the destination owner ATA-aware: when the payment ATA is created in this
+                // same transaction or bundle it has no pre-state, so fall back to the ATA-creation
+                // instruction (the same helper payment validation uses) instead of treating the
+                // transfer as a non-payment.
+                let destination_owner =
+                    match CacheUtil::get_account(config, rpc_client, destination_address, true)
```
amilz: i think you can use `TokenUtil::resolve_token_account_owner_and_mint` (`token/token.rs:774`)

crates/lib/src/validator/config_validator.rs
dev-jodee: Good catch — fixed in 0f5ab1e9c: the test now also removes JUPITER_API_KEY so it doesn't leak into other serial tests.

crates/lib/src/fee/fee.rs
dev-jodee: Real gap, but the fix needs a two-pass restructure (helper.rs accumulates all_bundle_instructions incrementally *after* each per-tx estimate, so later legs aren't visible at estimate time) and it only affects the bundle quote — phase-2 validation already does cross-tx ATA lookup, so enforcement is correct. Tracked as a follow-up: #604. Keeping this PR single-tx scoped.

crates/lib/src/fee/fee.rs
dev-jodee: Done in 0f5ab1e9c — analyze_payment_instructions now calls TokenUtil::resolve_token_account_owner_and_mint (made pub(crate)) instead of the inline duplicate, so the estimator and payment validation share the exact resolution path. The affected fee tests moved from mockall CacheUtil mocking to RPC-level account mocking, since the helper lives in token.rs which uses the real CacheUtil.

amilz (approved):
Fee estimator recognizes in-transaction payment ATAs via shared owner-resolution helper, config validation resolves auth secrets env-first to match runtime (no secret contents logged), and usage-limit window timestamp refreshed at store time to prevent stale-timestamp bypass. All 23 CI checks green.

PR #606, opened 2026-07-13
fix(fee): bundle-aware fee estimation for cross-leg ATA creation

### LINKED ISSUE #604: Bundle-aware fee estimation for cross-leg payment ATA creation
## Context
`FeeConfigUtil::analyze_payment_instructions` (fee estimation) only sees the current transaction's instructions. In a bundle, a payment whose destination ATA is created in one leg and paid in another isn't recognized by the per-tx estimate, so the quote can underprice it.

Payment **validation** already handles this: `calculate_payment_lamport_totals` takes `bundle_instructions` and does cross-tx ATA lookup (`bundle/helper.rs` phase 2). Only the estimator (phase 1) is not bundle-aware.

## Why it's non-trivial
`bundle/helper.rs` accumulates `all_bundle_instructions` incrementally as it loops, *after* each per-tx `estimate_kora_fee` call — so at estimate time for leg *i*, later legs aren't visible. Making phase-1 estimation bundle-complete needs a two-pass restructure (gather all instructions first, then estimate).

## Impact
Quote accuracy only — enforcement (phase 2) already validates cross-tx correctly, so no fund risk. A bundle paying via a cross-leg ATA-created destination may be underquoted.

## Scope
Follow-up to the single-tx fix (recognize in-transaction payment ATA creation in fee estimation). Raised in review of PR #603.

raushan728:
This PR makes the bundle fee estimator aware of ATA creation across legs. Before, Phase 1 estimation ran per‑transaction without seeing instructions from other legs, so a payment to an ATA created in an earlier leg was completely missed in the quote. Now the estimator receives the full bundle instructions (same context Phase 2 validation already uses). The missing‑payment penalty logic is also corrected so we don't overcharge when at least one leg does contain a valid payment.

Closes #604.

### DIFF
--- a/crates/lib/src/bundle/helper.rs
+++ b/crates/lib/src/bundle/helper.rs
@@ -146,14 +146,20 @@ impl BundleProcessor {
                     .run(&mut resolved_tx, config, rpc_client, &fee_payer, context)
                     .await?;
             }
+            all_bundle_instructions.extend(resolved_tx.all_instructions.clone());
+            resolved_transactions.push(resolved_tx);
+        }
 
+        // Calculate the estimated fee for each transaction using the full bundle context
+        for resolved_tx in resolved_transactions.iter_mut() {
             let fee_calc = FeeConfigUtil::estimate_kora_fee(
-                &mut resolved_tx,
+                resolved_tx,
                 &fee_payer,
                 config.validation.is_payment_required(),
                 rpc_client,
                 config,
                 transfer_hook_validation_flow,
+                Some(&all_bundle_instructions),
             )
             .await?;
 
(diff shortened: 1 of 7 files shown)


crates/lib/src/rpc_server/method/estimate_bundle_fee.rs
```diff
@@ -406,4 +407,107 @@ mod tests {
...
+            if *addr == payment_ata {
+                Err(KoraError::AccountNotFound(payment_ata.to_string()))
+            } else if *addr == mint {
+                Ok(mint_account_clone.clone())
+            } else if *addr == sender_ata {
+                Ok(sender_ata_account.clone())
+            } else {
+                Ok(Account::default())
+            }
+        });
```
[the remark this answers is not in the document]
raushan728: Opps! Leftover from a quick check to see which addresses the estimator was hitting during mock setup now its gone.

crates/lib/src/tests/config_mock.rs
```diff
@@ -172,6 +172,11 @@ impl ConfigMockBuilder {
         self
     }
 
+    pub fn with_price_model(mut self, price_model: crate::fee::price::PriceModel) -> Self {
```
dev-jodee: lets put the imports top of the file

crates/lib/src/fee/fee.rs
```diff
@@ -103,12 +104,15 @@ impl FeeConfigUtil {
...
         fee_payer: &Pubkey,
+        bundle_instructions: Option<&[Instruction]>,
     ) -> Result<(bool, u64), KoraError> {
         let payment_destination = config.kora.get_payment_address(fee_payer)?;
         let mut has_payment = false;
         let mut total_transfer_fees = 0u64;
 
-        let all_instructions = resolved_transaction.all_instructions.clone();
+        let all_instructions = bundle_instructions
+            .map(|instrs| instrs.to_vec())
```
dev-jodee: bundle_instructions.map(|instrs| instrs.to_vec()) clones the entire bundle instruction set on every leg's call, and it's only borrowed

crates/lib/src/bundle/helper.rs
```diff
@@ -163,18 +169,19 @@ impl BundleProcessor {
...
-        // If multiple transactions are missing payments, we've overcounted by
-        // (txs_missing_payment_count - 1) * ESTIMATED_LAMPORTS_FOR_PAYMENT_INSTRUCTION
-        if txs_missing_payment_count > 1 {
-            let overcount =
-                (txs_missing_payment_count - 1) * ESTIMATED_LAMPORTS_FOR_PAYMENT_INSTRUCTION;
+        // In a bundle, only one payment is required for the entire transaction sequence.
+        // We adjust the total fee to ensure we don't overcharge for missing payments.
+        let overcount = if (txs_missing_payment_count as usize) == resolved_transactions.len()
+            && txs_missing_payment_count > 0
+        {
```
dev-jodee, line 181: can we add one line comments explaining the if / else, what it means here

dev-jodee (changes requested):
couple of things, and please make the import at the top of the file for consistency

crates/lib/src/fee/fee.rs
raushan728: Switched to a borrowed slice, no more per-leg clone.

raushan728:
Review address in b4068f4cea1455364b84e9c331cd2b4344383947

dev-jodee (approved):

PR #610, opened 2026-07-14
Improve initialize-atas diagnostics for invalid and missing mints

### LINKED ISSUE #608: Devex: cross-cluster mint hint when initialize-atas fails on a missing or wrong-owner mint
## Background

`kora rpc initialize-atas` creates ATAs for every mint in `allowed_spl_paid_tokens`. For each mint, `find_missing_atas` (`crates/lib/src/admin/token_util.rs`) fetches the mint account and resolves its owning token program via `TokenType::get_token_program_from_owner` (`crates/lib/src/token/token.rs`).

A common setup mistake is configuring a mint that doesn't exist on the connected cluster — e.g. a mainnet USDC address while pointed at devnet, or a devnet mint on a fresh local validator. Solana Explorer handles this nicely by checking other clusters and telling you where the address actually lives. PR #464 brought that idea to Kora (`ConfigValidator::check_cross_cluster_mints`, behind `validation.cross_cluster_check` / `cross_cluster_endpoints`) — but only for `kora config validate-with-rpc`.

## Problem

When `initialize-atas` itself hits a bad mint, the user gets a raw error with no hint:

- Mint absent on cluster → `Failed to fetch mint account for {mint}: ...` (AccountNotFound).
- Address exists but isn't a mint (e.g. a wallet address, or an address that's a system account on this cluster) → `Invalid token program owner: 11111111111111111111111111111111`.

Neither tells the user the likely cause: wrong cluster for this mint.

## Proposed solution

Reuse the cross-cluster probe from #464 in the `initialize-atas` failure path:

- When the mint account fetch fails with account-not-found, run `check_cross_cluster_mints` against the configured (or default mainnet/devnet/testnet) endpoints and append a hint to the error: `mint {mint} not found on this cluster, but exists on mainnet-beta — check your --rpc-url`.
- When the owner is invalid, say what the account actually is instead of only printing the owner pubkey: `{mint} exists on this cluster but is not an SPL/Token-2022 mint (owner: System Program)`.

## Expected work

- [ ] Extract the cross-cluster probe so it's callable outside `ConfigValidator` (it currently lives in `crates/lib/src/validator/config_validator.rs`).
- [ ] Wire it into the mint-fetch error path of `find_missing_atas` in `crates/lib/src/admin/token_util.rs`, gated on the same `validation.cross_cluster_check` config.
- [ ] Improve the `Invalid token program owner` message in `get_token_program_from_owner` call sites to name known owners (System Program, etc.).
- [ ] Unit tests with mocked RPC: missing mint produces the cross-cluster hint; wrong-owner mint produces the descriptive message.
- [ ] Keep the happy path untouched — no extra RPC calls when all mints resolve.

## Acceptance criteria

- Running `initialize-atas` against the wrong cluster for a configured mint fails with an actionable message naming the cluster(s) where the mint exists.
- No cross-cluster probing occurs unless a mint lookup fails.

## References

- Cross-cluster validation for `validate-with-rpc`: #464

raushan728:
Improve `initialize-atas` diagnostics for missing and invalid mints.

- Add cross-cluster hints for missing mints.
- Use human-readable program owner names.
- Avoid a redundant local RPC during cross-cluster probing.

Closes #608

### DIFF
--- a/crates/lib/src/admin/token_util.rs
+++ b/crates/lib/src/admin/token_util.rs
@@ -4,7 +4,7 @@ use crate::{
     state::{get_request_signer_with_signer_key, get_signer_pool},
     token::token::TokenType,
     transaction::TransactionUtil,
-    validator::cross_cluster::check_cross_cluster_mints,
+    validator::cross_cluster::probe_missing_mints,
 };
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_compute_budget_interface::ComputeBudgetInstruction;
@@ -287,8 +287,7 @@ pub async fn find_missing_atas(
                     format!("Failed to fetch mint account for {mint}: account not found");
                 if config.validation.cross_cluster_check {
                     let mut warnings = Vec::new();
-                    check_cross_cluster_mints(
-                        rpc_client,
+                    probe_missing_mints(
                         &[mint.to_string()],
                         &config.validation.cross_cluster_endpoints,
                         &mut warnings,
--- a/crates/lib/src/admin/token_util.rs
+++ b/crates/lib/src/admin/token_util.rs
@@ -490,7 +490,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_find_missing_atas_cross_cluster_probe_success() {
+    async fn test_find_missing_atas_cross_cluster_probe_finds_mint() {
         let mint = Pubkey::new_unique();
 
         let mut server = mockito::Server::new_async().await;
--- a/crates/lib/src/admin/token_util.rs
+++ b/crates/lib/src/admin/token_util.rs
@@ -4,6 +4,7 @@ use crate::{
     state::{get_request_signer_with_signer_key, get_signer_pool},
     token::token::TokenType,
     transaction::TransactionUtil,
+    validator::cross_cluster::check_cross_cluster_mints,
 };
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_compute_budget_interface::ComputeBudgetInstruction;
@@ -32,6 +33,7 @@ This funciton is tested via the makefile, as it's a CLI command and requires a v
 
 const DEFAULT_CHUNK_SIZE: usize = 10;
 
+#[derive(Debug)]
 pub struct ATAToCreate {
     pub mint: Pubkey,
     pub ata: Pubkey,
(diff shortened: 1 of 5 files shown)


crates/lib/src/admin/token_util.rs
```diff
@@ -278,12 +280,52 @@ pub async fn find_missing_atas(
...
+                    format!("Failed to fetch mint account for {mint}: account not found");
+                if config.validation.cross_cluster_check {
+                    let mut warnings = Vec::new();
+                    check_cross_cluster_mints(
+                        rpc_client,
+                        &[mint.to_string()],
+                        &config.validation.cross_cluster_endpoints,
+                        &mut warnings,
+                    )
+                    .await;
```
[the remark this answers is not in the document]
raushan728: Addressed in 1de95c068a43dd5973ba58cdc2422abe5ab1787d I extracted the probing logic into `probe_missing_mints`, so `find_missing_atas` now reuses the cross-cluster probing without performing the redundant local `get_multiple_accounts` call

dev-jodee (approved):

crates/lib/src/admin/token_util.rs
```diff
@@ -449,6 +489,177 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_find_missing_atas_cross_cluster_probe_success() {
```
dev-jodee: test name says success, but the assert is on is_err

crates/lib/src/admin/token_util.rs
```diff
@@ -32,6 +33,7 @@ This funciton is tested via the makefile, as it's a CLI command and requires a v
 
 const DEFAULT_CHUNK_SIZE: usize = 10;
 
+#[derive(Debug)]
```
dev-jodee, line 36: Don't think you need debug since you never print it ?

crates/lib/src/validator/cross_cluster.rs
```diff
@@ -0,0 +1,212 @@
...
+    probe_missing_mints(&missing, endpoints, warnings).await;
+}
+
+pub(crate) async fn probe_missing_mints(
+    missing: &[String],
+    endpoints: &[String],
+    warnings: &mut Vec<String>,
+) {
+    // No probes can be performed without configured endpoints.
+    if missing.is_empty() || endpoints.is_empty() {
```
dev-jodee: if `cross_cluster_check` is true, but endpoints is empty we should warn

dev-jodee (changes requested):

crates/lib/src/admin/token_util.rs
raushan728: It is required here because the tests use `Result::unwrap_err`, which requires `T: Debug` `(Vec<ATAToCreate>` in this case). Removing the derive causes the tests to fail to compile.

dev-jodee (changes requested):
looks good but CI failing

raushan728:
> looks good but CI failing

Could you please re-runs the jobs?

raushan728:
There is no problem here. we just needed to do rebase.

raushan728:
The test that failed in this PR was added in our #565. It is already merged, but its commit wasn't here, so I just rebased it because the tests are passing on my local 🙂

dev-jodee (approved):

PR #614, opened 2026-07-20
feat(kora-deploy): add resume capability and buffer cleanup on transient failures

raushan728:
Adds resume capability and automatic buffer cleanup to kora-deploy, persisting local state so deployments can recover from transient failures instead of leaving orphaned on-chain buffers.

- Resume interrupted deployments with `--resume`, guarded by SHA-256 hash verification and chunk bounds checking
- Buffers are closed automatically on failure by default; pass `--no-cleanup-on-failure` to leave the buffer open for a later `--resume`
- Detect already-succeeded deploys after a timeout by checking ProgramData before resubmitting, skipping this check entirely on fresh (non-resume) deploys
- Force buffer cleanup when the initial state save fails, since the buffer can't be recovered any other way in that case
- Refactor deploy() into smaller functions with a shared, parametrized cleanup_buffer! macro
- Add `.kora-deploy-state.json` to `.gitignore` since it contains raw keypair bytes
- Add `share_mode(0)` on the non-Unix save path as a partial mitigation (concurrency guard only, not an ACL restriction full Windows permission support is out of scope here)
- Add unit tests for the state-invariant checks (hash mismatch, chunk-count overflow) via a standalone `check_state_invariants()` helper
- Document the resume/recovery flow and new flags in the README

### DIFF
--- a/crates/kora-deploy/src/lib.rs
+++ b/crates/kora-deploy/src/lib.rs
@@ -1,6 +1,14 @@
 #![allow(deprecated)]
 
-use std::{path::Path, str::FromStr, sync::Arc, time::Duration};
+pub mod state;
+
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    str::FromStr,
+    sync::Arc,
+    time::Duration,
+};
 
 use anyhow::{anyhow, bail, Context, Result};
 use base64::{engine::general_purpose::STANDARD as B64, Engine as _};
@@ -9,6 +17,7 @@ use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_commitment_config::CommitmentConfig;
 use solana_loader_v3_interface::{instruction as loader_v3, state::UpgradeableLoaderState};
 use solana_sdk::{
+    hash::hash,
     instruction::{AccountMeta, Instruction},
     message::Message,
     pubkey::Pubkey,
@@ -16,6 +25,8 @@ use solana_sdk::{
     signer::Signer,
     transaction::Transaction,
 };
+
+use crate::state::DeployState;
 
 const WRITE_CHUNK_SIZE: usize = 900;
 const BPF_LOADER_UPGRADEABLE: Pubkey =
@@ -37,6 +48,9 @@ pub struct DeployConfig<'a> {
     /// deploy registry, allowing future upgrades signed by it. Without a wallet the program
     /// is immutable through the paymaster.
     pub wallet: Option<&'a Keypair>,
+    pub resume: bool,
+    pub cleanup_on_failure: bool,
+    pub state_path: PathBuf,
 }
 
 pub struct UpgradeConfig<'a> {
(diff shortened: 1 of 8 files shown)


raushan728:
@greptile re-review

raushan728:
@greptile re-review

raushan728:
@greptile re-review

raushan728:
@greptile re-review

raushan728:
@greptile re-review

dev-jodee (changes requested):
- should .gitignore `.kora-deploy-state.json`
- should prob have tests for  hash-mismatch detection, written_chunks > chunk_count corruption handling, skip-already-written-chunks, is_already_live short-circuit, etc.

crates/kora-deploy/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::OpenOptionsExt;
+            let mut options = fs::OpenOptions::new();
+            options.write(true).create(true).truncate(true).mode(0o600);
+            let mut file = options.open(&temp_path).with_context(|| {
+                format!("failed to open temp deploy state for writing: {}", temp_path.display())
+            })?;
+            use std::io::Write;
```
dev-jodee: imports should be a top of module not in the code

crates/kora-deploy/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+            file.write_all(data.as_bytes()).with_context(|| {
+                format!("failed to write temp deploy state to {}", temp_path.display())
+            })?;
+            file.sync_data().context("failed to sync temp deploy state to disk")?;
+        }
+        #[cfg(not(unix))]
+        {
+            let mut file = fs::File::create(&temp_path).with_context(|| {
```
dev-jodee: here you can't set the permissions like you do in the unix branch?

raushan728: Hmm..  but, we can't set POSIX-style file mode (0o600) since it doesn't have the same permission model as Unix.

What I added instead is `share_mode(0)` which prevents other processes from concurrently opening the file while it's being written but this is a concurrency guard, not an ACL restriction. It does NOT restrict which users/processes can read the file once it's closed, so it's a partial mitigation at best.

For a real fix we'd need proper ACL manipulation, which requires an externaldependency (`windows-acl` or raw `winapi` calls) to set a DACL that restricts read access to the file owner only. I intentionally left that out of this PR
since:
- This project's CI runs on ubuntu-latest there's no windows target
  currently tested/supported
- Adding a windows-specific ACL dependency for an untested platform felt like
  scope creep for this PR

If this sounds. We can file a separate issue to track proper windows ACL support if/when this tool needs to officially support windows.

raushan728:
> * should .gitignore `.kora-deploy-state.json`

Added

> * should prob have tests for  hash-mismatch detection, written_chunks > chunk_count corruption handling, skip-already-written-chunks, is_already_live short-circuit, etc.

Added pure-logic unit tests for the two state-invariant checks

For the remaining cases you mentioned (skip-already-written-chunks, is_already_live short-circuit) agreed these need coverage too, but they're integration-style paths that depend on RPC/network calls (Kora + Solana RPC), which this crate doesn't currently have mocking infra for.

Rather than bolting on ad-hoc mocks in an already-large PR, we can file a follow-up issue to properly set up mock-based integration testing for kora-deploy. I'll scope that issue to cover both the integration tests AND adding a Windows CI target, since the non-Unix file-permission gap (raised above) would benefit from actually being exercised in CI rather than just reviewed by eye. That way both gaps get closed together with proper CI coverage backing them, instead of being two separate half-measures.

raushan728:
Addresses in 87743c22f1ed14f5b99209511ee5d570c245bedc

crates/kora-deploy/src/state.rs
```diff
@@ -39,18 +39,25 @@ impl DeployState {
...
         }
         #[cfg(not(unix))]
         {
-            let mut file = fs::File::create(&temp_path).with_context(|| {
-                format!("failed to create temp deploy state: {}", temp_path.display())
-            })?;
-            use std::io::Write;
+            // share_mode(0) limits concurrent access, but does not enforce
+            // file-level ACL restrictions (unlike Unix 0o600).
+            use std::os::windows::fs::OpenOptionsExt;
```
dev-jodee: imports should be top of mod

dev-jodee (commented):
all looks good except the 1 nit

crates/kora-deploy/src/state.rs
raushan728: I thought it just meant using it within that branch, so I wrote it inside anyway, I've done it

dev-jodee:
@raushan728 some CI failures

raushan728:
> @raushan728 some CI failures

Fixed in 902078118997ab0978b7cbd85088ef4d842b93cf

crates/kora-deploy/src/lib.rs
```diff
@@ -482,6 +482,7 @@ async fn finalize_deploy(
     Ok(())
 }
 
+// TODO: upgrade() has no state persistence, resume support, or buffer cleanup on failure.
```
raushan728, line 485: Will follow-up

crates/kora-deploy/src/state.rs
```diff
@@ -90,7 +92,8 @@ mod tests {
             program_hash: "dummyhash123".to_string(),
         };
 
-        let temp_file = temp_dir().join(format!("kora-deploy-state-test-{}.json", process::id()));
+        let temp_file =
+            temp_dir().join(format!("kora-deploy-state-test-{}.json", Pubkey::new_unique()));
```
raushan728: `Pubkey::new_unique()` uses an atomic counter internally, so it's unique across parallel tests in the same process just for future

dev-jodee (approved):

PR #626, opened 2026-07-30
ci: run fuzz targets in CI

dev-jodee:
## Summary
Adds `.github/workflows/fuzz.yml` so the cargo-fuzz targets run automatically instead of only locally.

- **smoke** (every PR/push touching `crates/**`, `fuzz/**`, `Cargo.*`, or the workflow): `cargo fuzz build` + a 60s run per target (`parse_transaction`, `decode_b64_transaction`). Fails on a build break **or** a crash, and uploads any reproducer as an artifact.
- **deep** (nightly cron `0 7 * * *` + manual `workflow_dispatch`): 10 min per target via a matrix, uploads the corpus and any crash artifacts.
- Uses the repo's existing conventions: `dtolnay/rust-toolchain@master` pinned to `nightly-2026-07-13` (cargo-fuzz needs nightly), `taiki-e/install-action` for a prebuilt cargo-fuzz binary, and `Swatinem/rust-cache` scoped to the `fuzz` workspace.

The `cargo fuzz build` step doubles as a guard that nothing currently has: if a kora-lib parser/validator API change breaks the fuzz harness, CI now catches it (unit tests don't compile the fuzz crate).

## Why a bounded PR smoke + nightly split
A short bounded PR run catches obvious regressions cheaply; deep bug-finding runs on the nightly schedule, off the PR critical path, so a latent crash surfaces as a nightly artifact rather than blocking an unrelated PR.

## Test Plan
- `actionlint` passes clean.
- The workflow's path filter includes `.github/workflows/fuzz.yml`, so the **smoke job runs on this PR** and self-validates the build + 60s run end to end. (First run builds cold under the sanitizer — hence the 45-min smoke timeout; steady-state is fast via rust-cache.)
- Locally: `cargo fuzz build` + `cargo fuzz run <target> -- -max_total_time=60` both succeed with no crashes.

## Follow-ups (not here)
- Commit/persist a seed corpus so runs accumulate coverage across nightlies (currently each run starts cold; the nightly uploads its discovered corpus as an artifact).

Closes TOO-625

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -44,6 +44,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly-2026-07-13
+          targets: x86_64-unknown-linux-gnu
       - uses: taiki-e/install-action@v2.82.0
         with:
           tool: cargo-fuzz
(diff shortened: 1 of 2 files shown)


amilz (approved):
CI-only: adds fuzz.yml (60s smoke per PR, 10m deep nightly) + README docs. Least-privilege perms, fuzz smoke passing on the PR. LGTM

PR #627, opened 2026-07-31
feat: add support for v1 transactions

dev-jodee:
## Summary
- Adds `VersionedMessage::V1` support end to end: decode/encode/sign/send/simulate all go through wincode, the wire codec for v1 (byte-identical to bincode for legacy/v0, so existing traffic is unchanged)
- Upgrades the solana crate stack 3.x → 4.x. The lockfile deliberately holds the June release wave: stable `solana-transaction-status-client-types` 4.1.2 is wincode 0.5 while the newest base crates moved to 0.6, so `Cargo.toml` upper-bounds `solana-message`/`solana-transaction`/`solana-program` until the stable client stack catches up (dependabot full-updates will fail loudly until then — expected)
- Temporary raw send/simulate helpers bypass the rpc-client's serde paths, which produce invalid v1 wire bytes; agave master already fixed this (anza-xyz/agave#12977) and the helpers get deleted once it ships stable
- v1 has no address lookup tables (resolution passes through) and ignores ComputeBudget instructions (config mask instead); size limits are version-aware (4096 bytes for v1, enforced at decode and in the Lighthouse size check)
- Test infra: `TransactionBuilder::v1()` with an explicit `TransactionConfig` (an empty mask requests 0 CUs and a 32KiB loaded-accounts cap — v1 clients must set limits), fuzz targets moved to wincode so v1 layouts are fuzzed

## Test Plan
- 914 unit tests pass (7 new: v1 decode roundtrip + signature over wire bytes, oversized rejection, per-version size limits, Lighthouse v1 append incl. a >1232-byte case, resolve, fee)
- Two new e2e tests verified against an agave 4.2.0-rc.0 test validator: `signTransaction` and `signAndSendTransaction` (respond_after=Confirmed) with v1 transactions — the confirmed send proves on-chain landing with USDC payment enforced
- The e2e tests self-skip on validators older than agave 4.2 (CI installs 4.0.3, which predates the `enable_tx_v1` gate) and activate automatically when CI's validator is bumped
- Full integration suite: 11/13 phases pass locally; the 2 failures are environmental and pre-existing (External Live = Jupiter API; loader_v3_deploy breaks on agave 4.2 due to an upstream gate — needs its own ticket, tracked in TOO-527)

## Breaking Changes
- None for the RPC interface or existing legacy/v0 traffic. Operators: no config changes required.

Follow-ups for prod readiness are tracked as a checklist in TOO-527 (kept open intentionally, hence Refs).

Refs: TOO-527

### DIFF
--- a/crates/lib/src/lighthouse/assertion.rs
+++ b/crates/lib/src/lighthouse/assertion.rs
@@ -7,10 +7,8 @@ use solana_sdk::{
 };
 
 use crate::{
-    config::LighthouseConfig,
-    constant::{LIGHTHOUSE_PROGRAM_ID, MAX_TRANSACTION_SIZE},
-    error::KoraError,
-    sanitize_error,
+    config::LighthouseConfig, constant::LIGHTHOUSE_PROGRAM_ID, error::KoraError, sanitize_error,
+    transaction::TransactionUtil,
 };
 
 /// Lighthouse instruction discriminators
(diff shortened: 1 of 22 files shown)


crates/lib/src/transaction/transaction.rs
```diff
@@ -15,8 +19,16 @@ impl TransactionUtil {
             KoraError::InvalidTransaction(format!("Failed to decode base64 transaction: {e}"))
         })?;
 
-        // First try to deserialize as VersionedTransaction
-        if let Ok(versioned_tx) = bincode::deserialize::<VersionedTransaction>(&decoded) {
+        // First try to deserialize as VersionedTransaction (wincode is the wire
+        // codec: byte-identical to bincode for legacy/v0, required for v1)
+        if let Ok(versioned_tx) = wincode::deserialize::<VersionedTransaction>(&decoded) {
```
amilz: wdyt about doing a match here to define a `transaction` that you could `sanitize`

```rust
        let transaction = match wincode::deserialize_exact::<VersionedTransaction>(&decoded) {
            Ok(versioned_tx) => versioned_tx,
            Err(_) => {
                let legacy_tx: Transaction = bincode::deserialize(&decoded).map_err(|e| {
                    KoraError::InvalidTransaction(format!("Failed to deserialize transaction: {e}"))
                })?;
                VersionedTransaction {
                    signatures: legacy_tx.signatures,
                    message: VersionedMessage::Legacy(legacy_tx.message),
                }
            }
        };

        transaction.message.sanitize().map_err(|e| {
            KoraError::InvalidTransaction(format!("Invalid transaction message: {e}"))
        })?;

      // and maybe add a size check?
```

crates/lib/src/lighthouse/assertion.rs
```diff
@@ -267,25 +266,23 @@ impl LighthouseUtil {
         let mut tx_with_assertion = transaction.clone();
         Self::append_instruction_to_transaction(&mut tx_with_assertion, assertion_ix)?;
 
-        let new_size = bincode::serialize(&tx_with_assertion)
+        let new_size = TransactionUtil::serialize_versioned_transaction(&tx_with_assertion)
```
amilz, line 269: want to do a `sanitize` before doing all this

amilz (commented):
lgtm. couple of suggestions. will ping you offline w/ a couple questions

crates/lib/src/lighthouse/assertion.rs
dev-jodee: superseded with the first fix (onyour comment)

amilz (dismissed):

amilz (approved):

PR #629, opened 2026-08-04
ci: baseline inactive rkyv advisory

dev-jodee:
## Summary
- baseline `RUSTSEC-2026-0235`, which `cargo audit` finds through `rust_decimal` lockfile metadata even though Kora does not enable its optional `rkyv` features
- document why the advisory does not affect the active dependency graph
- remove the obsolete `RUSTSEC-2026-0204` exception now that `crossbeam-epoch` is patched at 0.9.20

Upstream tracking: https://github.com/paupino/rust-decimal/issues/818

## Test Plan
- `cargo audit`
- `cargo tree -i rkyv@0.7.46` (no active reverse dependencies)

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -8,7 +8,7 @@ ignore = [
     "RUSTSEC-2026-0098",
     # rustls-webpki 0.101.x is pinned by the transitive rustls 0.21 TLS stack.
     "RUSTSEC-2026-0099",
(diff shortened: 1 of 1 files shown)


amilz (approved):
cargo audit baseline: swaps obsolete RUSTSEC-2026-0204 exception for documented inactive rkyv advisory (RUSTSEC-2026-0235). All CI green.

PR #631, opened 2026-08-05
chore: release sdk 3

amilz:
Release includes latest kit & kit plugin deps so this works w/ latest kit stack.


amilz:
Superseded — reopened from a topic branch to satisfy the mainline PR policy check.

PR #635, opened 2026-08-14
refactor(sdk-ts): return execution context from Kora transaction plan executor

amilz:
## Summary

- Migrates `createKoraTransactionPlanExecutor` off a return form that [`@solana/kit` v7.1.0](https://github.com/anza-xyz/kit/releases/tag/v7.1.0) deprecates ([#1915](https://github.com/anza-xyz/kit/pull/1915)). `executeTransactionMessage` now returns the success **context** rather than a bare `Signature`.
- Bumps the `@solana/kit` peer dependency `^7.0.0` → `^7.1.0` and refreshes `sdks/pnpm-lock.yaml`. This is **required, not cosmetic**: 7.0.0's `ExecuteTransactionMessage` type accepts only `Signature | Transaction`, so the migration does not typecheck against it.

```diff
  if (result.signature) {
-     return signature(result.signature);
+     return { signature: signature(result.signature) };
  }
  const signedTxBytes = getBase64Encoder().encode(result.signed_transaction);
  const decodedTx = getTransactionDecoder().decode(signedTxBytes);
- return getSignatureFromTransaction(decodedTx);
+ return { signature: getSignatureFromTransaction(decodedTx), transaction: decodedTx };
```

On the second path the decoded transaction is returned alongside the signature, so callers see it on the result context instead of only its signature.

Kora is exactly the case the release note calls out: an executor that deliberately produces partially signed transactions, to be paid for and submitted by a relayer later, can now succeed by returning its own signature alongside the transaction. A returned context must carry a `signature` — that is how the executor distinguishes a context from a `Transaction`, which keeps signatures in a `signatures` map and so never has one.

The mutable `_context` argument is deliberately left untouched. On the failure path the executor derives a signature from any `transaction` left on the context, which for a fee-payer-unsigned transaction would throw `SOLANA_ERROR__TRANSACTION__FEE_PAYER_SIGNATURE_MISSING` and mask the real error.

## Test plan

- [x] `just build` — Rust workspace compiles clean
- [x] `just unit-test-ts` — 3 suites, **87 passed** / 1 skipped
- [x] `pnpm run type-check` in `sdks/ts` — clean against real 7.1.0 types
- [x] Verified `sdks/ts/test/kit-client.test.ts`, which asserts on `result.context.signature`, still passes — the returned context and the mutable context merge, returned value winning

## ⚠️ CI will fail until 2026-08-15 ~13:44 UTC — please do not merge before then

pnpm 10.29 enforces a 24-hour `minimumReleaseAge`. The `@solana/*` 7.1.0 packages were published 2026-08-14 13:36–13:44 UTC, so a lockfile pinning them fails `pnpm install --frozen-lockfile` with `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` (43 entries). This is purely a timing artifact and clears itself once the packages are 24h old; re-run CI after that and it goes green with no code change.

Note for reviewers: a local `pnpm install` will try to "fix" this by auto-writing a 43-package `minimumReleaseAgeExclude` list into `sdks/pnpm-workspace.yaml` (and dropping `auditConfig: {}`). That was deliberately reverted out of this PR — it bypasses the supply-chain guard rather than waiting for it, and it is not this repo's policy. Please don't commit it. For the same reason the commit skipped the `format-ts-sdk` pre-commit hook, which runs `pnpm install` and cannot pass today.

Unrelated pre-existing issue spotted: `pnpm run lint` in `sdks/ts` crashes with `TypeError: context.getSourceCode is not a function` in `eslint-plugin-typescript-sort-keys` under ESLint 10.6.0. Confirmed identical on a clean `origin/main` worktree — not caused by this change.

Part of DEV-896.

### DIFF
--- a/sdks/pnpm-lock.yaml
+++ b/sdks/pnpm-lock.yaml
@@ -25,35 +25,35 @@ importers:
     dependencies:
       '@solana-program/compute-budget':
         specifier: ^0.17.0
-        version: 0.17.0(@solana/kit@7.0.0(typescript@6.0.3))
+        version: 0.17.0(@solana/kit@7.1.0(typescript@6.0.3))
       '@solana-program/token':
         specifier: ^0.15.0
-        version: 0.15.0(@solana/kit@7.0.0(typescript@6.0.3))
+        version: 0.15.0(@solana/kit@7.1.0(typescript@6.0.3))
       '@solana/kit':
-        specifier: ^7.0.0
-        version: 7.0.0(typescript@6.0.3)
+        specifier: ^7.1.0
+        version: 7.1.0(typescript@6.0.3)
       '@solana/kit-plugin-instruction-plan':
         specifier: ^0.13.0
-        version: 0.13.0(@solana/kit@7.0.0(typescript@6.0.3))
+        version: 0.13.0(@solana/kit@7.1.0(typescript@6.0.3))
       '@solana/kit-plugin-rpc':
         specifier: ^0.15.0
-        version: 0.15.0(@solana/kit@7.0.0(typescript@6.0.3))
+        version: 0.15.0(@solana/kit@7.1.0(typescript@6.0.3))
       '@solana/kit-plugin-signer':
         specifier: ^0.13.0
-        version: 0.13.0(@solana/kit@7.0.0(typescript@6.0.3))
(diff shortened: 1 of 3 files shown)


dev-jodee (approved):
kit v7.1.0 migration: execution context return. Tests pass, CI green.

PR #638, opened 2026-08-18
feat: add max_priority_fee_lamports validation config

dev-jodee:
## Summary
- Adds `validation.max_priority_fee_lamports`: an explicit cap on the priority fee a transaction may request. Unset = unlimited (backward compatible), `0` = priority fees not allowed.
- Enforced uniformly in `TransactionValidator::validate_transaction`, per version from its native source: v1 reads `message.config.priority_fee` directly; legacy/v0 parses `SetComputeUnitPrice`/`SetComputeUnitLimit` and computes ceil(price x limit / 1e6). When a price is set without an explicit limit, the 1.4M CU protocol maximum is assumed (an upper bound, so the cap can only over-reject).
- Motivation: operators could block priority fees for legacy/v0 by omitting the ComputeBudget program from `allowed_programs`, but v1 carries the priority fee in the message config rather than an instruction, so the program allowlist cannot gate it. Under `Free` or `Fixed { strict: false }` pricing this was an unintended fee-payer cost channel bounded only by `max_allowed_lamports`.
- Operator docs and commented `kora.toml` example included.

Note: the checked-in OpenAPI spec (`combined_api.json`) was not regenerated because the `docs` feature does not compile on main (pre-existing utoipa drift in `openapi/docs.rs`, 28 errors, unrelated to this change).

## Test Plan
- 10 new unit tests: priority fee extraction matrix (price+limit, ceil rounding, price-without-limit fallback, no/zero price, v1 config read, v1 ignoring ComputeBudget instructions) and validator enforcement (unset cap, over/under cap for both v1 config and ComputeBudget instructions, zero cap). Full kora-lib suite: 945 passed.
- 3 new integration tests against the live test validator (cap set to 5,000,000 in kora-test.toml): over-cap v1 config fee and over-cap legacy ComputeBudget fee rejected at `signTransaction` with the exact error, under-cap v1 fee signs normally with payment covering base plus priority. Regular phase: rpc 62/62, tokens 17/17, adversarial 9/9.

## Breaking Changes
- None. The field defaults to unset, which preserves current behavior exactly.

Closes TOO-675

### DIFF
--- a/crates/lib/src/fee/fee.rs
+++ b/crates/lib/src/fee/fee.rs
@@ -810,17 +810,21 @@ impl TransactionFeeUtil {
     /// is set without an explicit limit, the 1.4M CU protocol maximum is assumed
     /// (an upper bound, so a configured cap can only over-reject, never let a
     /// higher fee through).
-    pub fn get_requested_priority_fee(message: &VersionedMessage) -> u64 {
+    ///
+    /// `account_keys` must be the fully resolved key list (static plus lookup-table
+    /// loaded, in canonical order): the runtime resolves ComputeBudget program ids
+    /// against loaded keys too, so a static-only lookup would miss an ALT-loaded
+    /// ComputeBudget program and undercount the fee.
+    pub fn get_requested_priority_fee(message: &VersionedMessage, account_keys: &[Pubkey]) -> u64 {
         if let VersionedMessage::V1(v1_message) = message {
             return v1_message.config.priority_fee.unwrap_or(0);
         }
 
         let compute_budget_program_id = solana_compute_budget_interface::id();
-        let static_keys = message.static_account_keys();
         let mut compute_unit_price: Option<u64> = None;
         let mut compute_unit_limit: Option<u32> = None;
         for instruction in message.instructions() {
-            if static_keys.get(instruction.program_id_index as usize)
+            if account_keys.get(instruction.program_id_index as usize)
                 != Some(&compute_budget_program_id)
             {
                 continue;
@@ -876,7 +880,9 @@ mod tests {
     use solana_address_lookup_table_interface::{
         instruction as alt_instruction, program::ID as ADDRESS_LOOKUP_TABLE_PROGRAM_ID,
     };
-    use solana_message::{v0, v1, Message, VersionedMessage};
+    use solana_message::{
+        compiled_instruction::CompiledInstruction, v0, v1, Message, VersionedMessage,
+    };
     use solana_sdk::{
         account::Account,
         hash::Hash,
@@ -2524,7 +2530,10 @@ mod tests {
             transfer(&payer, &recipient, 1_000),
         ]);
 
-        assert_eq!(TransactionFeeUtil::get_requested_priority_fee(&message), 7_500);
+        assert_eq!(
+            TransactionFeeUtil::get_requested_priority_fee(&message, message.static_account_keys()),
+            7_500
+        );
     }
 
     #[test]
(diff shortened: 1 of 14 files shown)


crates/lib/src/fee/fee.rs
```diff
@@ -800,6 +804,49 @@ impl TransactionFeeUtil {
...
+            return v1_message.config.priority_fee.unwrap_or(0);
+        }
+
+        let compute_budget_program_id = solana_compute_budget_interface::id();
+        let static_keys = message.static_account_keys();
+        let mut compute_unit_price: Option<u64> = None;
+        let mut compute_unit_limit: Option<u32> = None;
+        for instruction in message.instructions() {
+            if static_keys.get(instruction.program_id_index as usize)
+                != Some(&compute_budget_program_id)
```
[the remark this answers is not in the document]
dev-jodee: Valid catch, fixed in 7987aaa33. Verified against agave 4.2: `SanitizedMessage::program_instructions_iter` (which feeds `TransactionConfiguration` and thus the fee actually charged) resolves program ids against the full loaded key list, and programs may legitimately be loaded via lookup tables (only signers must be static). `get_requested_priority_fee` now takes the resolved account key list and `validate_priority_fee` passes `transaction_resolved.all_account_keys` (static + canonical ALT order, matching v0 index semantics). Top-level-instructions-only behavior is unchanged, matching the runtime. Two regression tests added: extraction with the ComputeBudget id past the static-key boundary, and validator rejection of an ALT-loaded 15k fee against a 10k cap.

crates/lib/src/fee/fee.rs
```diff
@@ -800,6 +802,51 @@ impl TransactionFeeUtil {
...
+        }
+
+        let Some(price) = compute_unit_price.filter(|price| *price > 0) else {
+            return 0;
+        };
+
+        ComputeBudgetLimits {
+            compute_unit_price: price,
+            compute_unit_limit: compute_unit_limit
+                .unwrap_or(MAX_COMPUTE_UNIT_LIMIT)
```
amilz: should we use the same math they use in agave for default value rather than max?
https://github.com/anza-xyz/agave/blob/master/compute-budget-instruction/src/compute_budget_instruction_details.rs#L196-L219

crates/lib/src/validator/transaction_validator.rs
```diff
@@ -1391,6 +1419,132 @@ mod tests {
...
+        let transaction = resolved_from_message(create_v1_message(
+            &fee_payer,
+            v1::TransactionConfig::empty().with_priority_fee(u64::MAX),
+        ));
+
+        assert!(validator.validate_priority_fee(&transaction).is_ok());
+    }
+
+    #[test]
+    fn test_validate_priority_fee_enforces_cap_for_v1_config() {
```
amilz, line 1467: i think probably worth adding a test showing v1 txn with compute instructions (that should get silently ignored)...e.g. i have fee cap of 0.1 SOL. -- my config shows 0.09 SOL but have a compute budget that adds up to 0.2 sol....should be allowed b/c teh compute budget ix is ignored. and maybe a flip case where my config shows 0.11 (not allowed) and compute adds up to 0.09...should get rejected.

amilz (commented):
Looks great. 2 suggestions

crates/lib/src/fee/fee.rs
dev-jodee: yeah agreed, but don't want to bring in the crate directly (lots of dependencies), but I did use their math

crates/lib/src/validator/transaction_validator.rs
dev-jodee: yess

amilz (approved):

PR #648, opened 2026-08-22
test(validator): cover ALT fee-payer roles

### LINKED ISSUE #641: Tests: Address Lookup Table fee-payer drain property coverage
### What would you like?

Add Address Lookup Table roles to the fee-payer drain property test, so the ALT policy matrix is machine-checked rather than relying on hand-written one-flag-at-a-time cases.

**Background.** The `[validation.fee_payer_policy.*]` flags in `kora.toml` decide when the fee payer (the paymaster's own key, which signs every transaction) may appear in a privileged slot of an instruction. Every flag defaults to `false`. The invariant under test:

> The fee payer is gated in a given role exactly when that role's flag is off, and a non-fee-payer is never gated.

**The harness already exists.** Add one file:

```
crates/lib/src/validator/transaction_validator/fee_payer_policy_props/alt.rs
```

implementing the `DrainRole` trait, and declare it with `mod alt;` in that directory's `mod.rs`. `system.rs` in the same directory is the reference implementation: copy its shape, including the `proptest!` block at the bottom.

```rust
impl DrainRole for AltRole {
    const ROLES: &'static [Self] = &[/* one variant per gated account slot */];
    fn allowed_programs() -> Vec<String> { /* ALT program id */ }
    fn flag(self, policy: &mut FeePayerPolicy) -> &mut bool { /* &mut policy.alt.* */ }
    fn instruction(self, actor: &Pubkey) -> Instruction { /* put `actor` in the gated slot */ }
}
```

**Flags to cover (5):** `allow_create`, `allow_extend`, `allow_freeze`, `allow_deactivate`, `allow_close`.

**Which slot the validator gates** (see `validate_fee_payer_usage` in `crates/lib/src/validator/transaction_validator.rs`):

| Instruction | Gated slots |
|---|---|
| CreateLookupTable | authority **or** payer |
| ExtendLookupTable | authority **or** payer (payer is optional) |
| FreezeLookupTable | authority |
| DeactivateLookupTable | authority |
| CloseLookupTable | authority |

Where a flag gates more than one slot, add one role variant per slot. `system.rs` does this with `CreateAccountPayer` / `CreateAccountNewAccount`, both mapping to the same flag.

### Acceptance criteria

- Every flag above has at least one role, and the two-slot flags have one role per slot.
- `cargo test -p kora-lib --lib fee_payer_policy_props` passes.
- `just check` is clean.
- **Mutation proof in the PR description.** Point one role's `flag()` at a different policy field, show the property failing, then revert. A property test that cannot fail is worse than no test, and this is the one thing review cannot easily verify by reading.

### Notes

- Instruction builders live in `solana_address_lookup_table_interface`, already used by the unit tests in `transaction_validator.rs`. `create_lookup_table` returns the instruction alongside the derived table address.
- ALT has no unconditional drainage guard, so every case reaches the flag gate. (Some other program types do have one, which is why their issues carry extra warnings.)
- **If your property fails against current code, that is a possible fee-payer drain gap, not a PR.** Do not post the reproducer, the failing seed, or the `proptest-regressions` file. Report it privately via the advisory link in `SECURITY.md`.
- Comment to claim. One program type per contributor, please, so the files stay independent.

### Area

Transaction handling

RlxChap2:
## Summary
- Add `DrainRole` coverage for all five Address Lookup Table fee-payer policy flags.
- Cover separate authority and payer slots for `CreateLookupTable` and `ExtendLookupTable`.
- Run the shared property with 512 generated cases.

## Test plan
- [x] `cargo test -p kora-lib --lib fee_payer_policy_props`
- [x] `just check`
- [x] `cargo test --lib --workspace --exclude tests --quiet`

## Mutation proof
As an artificial mutation, I temporarily mapped `FreezeAuthority::flag()` to `policy.alt.allow_deactivate`. The property failed with:

> FreezeAuthority by fee payer must pass when its flag is on

I restored the mapping to `policy.alt.allow_freeze`; the property passed.

## Notes
Test-only change. Production validator behavior is unchanged.

Closes #641

### DIFF
--- a/crates/lib/src/validator/transaction_validator/fee_payer_policy_props/alt.rs
+++ b/crates/lib/src/validator/transaction_validator/fee_payer_policy_props/alt.rs
@@ -1,0 +1,83 @@
+use super::{assert_role_gated_iff_flag_off, role_and_flags, DrainRole};
+use crate::config::FeePayerPolicy;
+use proptest::prelude::*;
+use solana_address_lookup_table_interface::{
+    instruction::{
+        close_lookup_table, create_lookup_table, deactivate_lookup_table, extend_lookup_table,
+        freeze_lookup_table,
+    },
+    program::ID as ADDRESS_LOOKUP_TABLE_PROGRAM_ID,
+};
+use solana_sdk::{instruction::Instruction, pubkey::Pubkey};
+
+#[derive(Debug, Clone, Copy)]
+enum AltRole {
+    CreateAuthority,
+    CreatePayer,
+    ExtendAuthority,
+    ExtendPayer,
+    FreezeAuthority,
+    DeactivateAuthority,
+    CloseAuthority,
+}
+
+impl DrainRole for AltRole {
+    const ROLES: &'static [Self] = &[
+        Self::CreateAuthority,
+        Self::CreatePayer,
+        Self::ExtendAuthority,
+        Self::ExtendPayer,
+        Self::FreezeAuthority,
+        Self::DeactivateAuthority,
+        Self::CloseAuthority,
+    ];
+
+    fn allowed_programs() -> Vec<String> {
+        vec![ADDRESS_LOOKUP_TABLE_PROGRAM_ID.to_string()]
+    }
+
+    fn flag(self, policy: &mut FeePayerPolicy) -> &mut bool {
+        match self {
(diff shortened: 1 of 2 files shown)


dev-jodee (approved):

dev-jodee:
thanks @RlxChap2

PR #652, opened 2026-08-24
chore: replace keychain-sync plugin with a skill

dev-jodee:
## Summary
- Replaces the `.claude-plugin/kora-keychain-sync` plugin (3 files) with `.claude/skills/sync-keychain/SKILL.md`. The plugin's only unique capability was auto-provisioning a `repomix` MCP server for one slash command; a shallow clone plus `grep` covers it, so the marketplace and plugin manifests go away.
- Fixes the enumeration bug: the old command found upstream backends by `from_*` method, which also matches key-format helpers (`from_bytes`, `from_pem`, `from_u8_array_string`, ...). It now enumerates by `*SignerConfig` struct, which returns the 14 real backends with no false positives.
- Fixes the stale variant list: the old Step 4 hardcoded Memory/Turnkey/Privy/Vault, so it would have re-scaffolded 8 signers that already exist. The list is now derived from the `SignerTypeConfig` enum at runtime.
- Scaffolding templates updated to the current adapter shape (build a `solana_keychain::XSignerConfig` and await, per `build_fireblocks_signer`), plus the two touchpoints the old command missed: the exhaustive `http_config` match in `signer_validator.rs` and `cargo update -p solana-keychain` after a bump.

## Test Plan
Every command in the skill was run against the repo:

| Step | Command | Result |
|---|---|---|
| 1 | crates.io version vs `crates/lib/Cargo.toml` | `current=1.4.0 latest=1.4.0` |
| 2 | clone + grep `*SignerConfig` upstream | 14 backends |
| 3 | `awk` over `SignerTypeConfig` | 12 variants |
| 5 | `cargo test -p kora-lib --lib signer -- --list` | 84 tests match |

The Step 2 / Step 3 delta is **Fordefi** and **Utila**, the two backends `solana-keychain` ships that Kora has no adapter for. Not added here; flagged for a follow-up decision on whether they are wanted.

Note for reviewers: the `git clone` in Step 2 needs the Claude Code sandbox disabled (`unable to get credential storage lock`). Left as a plain clone rather than encoding a local-environment workaround in the skill.


amilz (approved):
Replaces keychain-sync plugin with a skill; enumeration/stale-variant bugs fixed and each step verified against the repo. All checks green. LGTM.

PR #654, opened 2026-08-26
fix(transaction): reject v1 transactions that cannot execute

amilz:
- reject transactions that have v1 message configs that are unset
- lighthouse assert -- fail if sanitze has too many accounts

Refs: TOO-527

### DIFF
--- a/crates/lib/src/lighthouse/assertion.rs
+++ b/crates/lib/src/lighthouse/assertion.rs
@@ -274,16 +274,23 @@ impl LighthouseUtil {
             })?
             .len();
 
-        let max_size = TransactionUtil::max_transaction_size(&tx_with_assertion.message);
-        if new_size > max_size {
+        // Appending the assertion adds accounts and an instruction, either of which can push
+        // the message past a limit that is not about byte size
+        let rejection = if let Err(e) = tx_with_assertion.message.sanitize() {
+            Some(format!("would produce an invalid message ({e})"))
+        } else {
+            let max_size = TransactionUtil::max_transaction_size(&tx_with_assertion.message);
+            (new_size > max_size)
+                .then(|| format!("would exceed transaction size limit ({new_size} > {max_size})"))
+        };
+
+        if let Some(reason) = rejection {
             if config.fail_if_transaction_size_overflow {
                 return Err(KoraError::ValidationError(format!(
-                    "Adding Lighthouse assertion would exceed transaction size limit ({new_size} > {max_size})"
+                    "Adding Lighthouse assertion {reason}"
                 )));
             } else {
-                log::warn!(
-                    "Lighthouse assertion would exceed transaction size limit ({new_size} > {max_size}). Skipping."
-                );
+                log::warn!("Lighthouse assertion {reason}. Skipping.");
                 return Ok(());
             }
         }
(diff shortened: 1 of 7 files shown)


crates/lib/src/lighthouse/assertion.rs
```diff
@@ -483,6 +490,67 @@ mod tests {
         assert!(transaction.message.static_account_keys().contains(&LIGHTHOUSE_PROGRAM_ID));
     }
 
+    /// Build a V1 transaction whose account list already sits at the format's 64-address
+    /// limit, so appending any account the message does not already hold makes it
+    /// unsanitary.
+    fn v1_transaction_with_max_addresses(keypair: &Keypair) -> VersionedTransaction {
+        let mut account_keys = vec![keypair.pubkey()];
+        account_keys.extend((1..v1::MAX_ADDRESSES).map(|_| Pubkey::new_unique()));
```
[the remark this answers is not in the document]
amilz: Fixed in 481def52. The V1 fixtures now use fixed byte arrays: keypairs via `Keypair::new_from_array`, the filler addresses in the Lighthouse address-limit test derived from their index, and fixed blockhashes. None of these tests depend on the key values, so a failure now reproduces from the test alone.

The two decode tests also share one `encoded_v1_transaction_with_config` helper instead of repeating the setup.

Left the pre-existing `Keypair::new()` / `Pubkey::new_unique()` fixtures in the surrounding tests alone — changing those is a wider cleanup than this PR.

crates/lib/src/lighthouse/assertion.rs
```diff
@@ -483,6 +490,68 @@ mod tests {
         assert!(transaction.message.static_account_keys().contains(&LIGHTHOUSE_PROGRAM_ID));
     }
 
+    /// Build a V1 transaction whose account list already sits at the format's 64-address
+    /// limit, so appending any account the message does not already hold makes it
+    /// unsanitary.
+    fn v1_transaction_with_max_addresses(keypair: &Keypair) -> VersionedTransaction {
```
dev-jodee: can we have utils functions in our test utils files

crates/lib/src/transaction/transaction.rs
```diff
@@ -48,6 +50,41 @@ impl TransactionUtil {
...
+        }
+
+        if !unset_limits.is_empty() {
+            return Err(KoraError::InvalidTransaction(format!(
+                "V1 transaction config must set {} to a non-zero value",
+                unset_limits.join(" and ")
+            )));
+        }
+
+        if v1_message.account_keys.contains(&solana_compute_budget_interface::id()) {
```
dev-jodee: this should go in transaction validator, the rest is good here tho

dev-jodee (changes requested):

crates/lib/src/lighthouse/assertion.rs
amilz: [8032907](https://github.com/solana-foundation/kora/pull/654/commits/80329070abaefe1ce9f89a443b53c61fd4feac26)

crates/lib/src/transaction/transaction.rs
amilz: [8032907](https://github.com/solana-foundation/kora/pull/654/commits/80329070abaefe1ce9f89a443b53c61fd4feac26)

dev-jodee (approved):

PR #656, opened 2026-08-27
docs: slim skills to guides and push detail into references

dev-jodee:
## Summary

Applies two rules to the five skills in `.claude/skills/`: keep them lightweight guides that help Claude *find* information rather than prescribe every step, and use progressive disclosure so long content lives in separate files.

SKILL.md bodies load in full on every trigger; `references/` load on demand. That budget went **893 → 365 lines**, with detail relocated rather than deleted (3 new reference files).

| Skill | SKILL.md | Moved to references |
|---|---|---|
| kora-operator | 250 → 99 | new `deployment.md` |
| kora-client | 186 → 82 | worked example already existed in `guides.md` |
| sync-keychain | 177 → 69 | new `scaffolding.md` (5 code templates) |
| release | 167 → 73 | new `changelog.md` (git-cliff branching) |
| complete-release | 113 → 42 | — |

What got cut is copy-paste bash for routine steps (`git status`, `checkout -b`, `npm version`, `gh pr create` boilerplate) and reference material duplicated from the `references/` files sitting next to it. What got kept, and in places sharpened, is the non-obvious constraints — the "highly important areas" where being prescriptive earns its cost.

## Two defects found while auditing

**1. `release` would publish the wrong crate.** It instructed `cargo set-version --workspace`. The workspace includes `crates/kora-deploy`, versioned independently at `0.2.0` and with no `publish = false`. A release at 2.2.x would bump it to 2.2.x and CI would push it to crates.io. Replaced with `cargo set-version -p kora-lib -p kora-cli` plus the two explicit root `Cargo.toml` edits (`[workspace.package]`, the `kora-lib` pin), and a verification step asserting `crates/kora-deploy/Cargo.toml` is unchanged.

**2. `kora-operator` understated the Lighthouse gap.** It said Lighthouse "only works with signTransaction/signBundle". It does not merely not-apply on broadcast methods — `LighthouseUtil::add_fee_payer_assertion` returns early on `will_send` (`lighthouse/assertion.rs:43`), so an operator who enables it for drain protection on a `signAndSend*` node gets none, silently. `config_validator.rs:557` warns but does not block startup. Now stated as the failure mode.

## Also fixed: stale config reference

`kora-operator/references/configuration.md` enumerated the fee payer policy fields and had drifted: 12 `spl_token` flags listed against 14 in `config.rs`, `token_2022` annotated "same 12 fields" against its actual 16, and the `alt` / `bpf_loader_upgradeable` / `loader_v4` sections missing entirely. Replaced the enumeration with the section list, the naming rule, and a pointer to `config.rs` + `kora config validate` — the list regrows every time an instruction is gated, so restating it guarantees the next drift.

## Test Plan
- Docs only; no Rust or TS touched.
- All 11 `references/*.md` links from SKILL.md files resolve (verified).
- Claims re-checked against source: `Cargo.toml` workspace members and per-crate versions, `assertion.rs` early return, `SplTokenInstructionPolicy`/`Token2022InstructionPolicy` field counts, `fallback_if_unavailable` semantics.

## Notes for the reviewer
- The `release` bump is the one part worth reading closely — it is the only behavioural change, and a wrong version bump is expensive. Worth a dry run on a throwaway branch before the next release.
- Stacked conceptually on #655 (CLAUDE.md), but independent: no file overlap, either can merge first.

### DIFF
--- a/.claude/skills/complete-release/SKILL.md
+++ b/.claude/skills/complete-release/SKILL.md
@@ -5,109 +5,38 @@ description: "Reviewer workflow for completing a Kora mainline release after the
 
 # Kora Complete Release
 
-Run this after a mainline release PR has been approved. It squash-merges the PR and triggers the appropriate CI publish workflows on `main`.
+Run after a mainline release PR is approved: squash-merge it, then trigger the publish workflows
+on `main`.
 
----
+## Merge
 
-## Step 1 — Identify the release PR
+Find the release PR (`gh pr list --base main --search "chore: release"`) and confirm the number
+with the user before merging — this publishes to crates.io and npm, which cannot be undone.
+Squash-merge, delete the branch, then pull `main`.
 
-```bash
-# List open release PRs
-gh pr list --base main --search "chore: release" --json number,title,headRefName
-```
+## Decide which workflows to run
 
-Confirm the correct PR number with the user before proceeding.
-
----
-
-## Step 2 — Squash-merge the PR
+Publish only what the release actually changed. Inspect the merge commit: paths under
+`crates/`, `Cargo.toml`, `Cargo.lock` or `CHANGELOG.md` mean Rust; paths under `sdks/` mean
+TypeScript. A combined release triggers both.
 
 ```bash
-gh pr merge "${PR_NUMBER}" --squash --delete-branch
+gh workflow run "Publish Rust Crates" --ref main \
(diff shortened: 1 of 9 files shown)


dev-jodee:
Folded into #655 — same commit, one PR for the whole docs pass.

PR #658, opened 2026-09-01
feat(rpc): make `CORS` allow `origins` configurable

raushan728:
Replaces the hardcoded `allow_origin(Any)` CORS policy with a configurable `cors_allow_origins` field, allowing operators to restrict cross-origin access in production deployments instead of allowing all origins by default.

Note: this is a split-out of #559.

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -26,6 +26,67 @@ use crate::{
 
 pub use crate::usage_limit::{UsageLimitConfig, UsageLimitRuleConfig};
 
+/// Validates a CORS origin string
+pub fn is_valid_cors_origin(origin: &str) -> bool {
+    if origin == "*" {
+        return true;
+    }
+
+    let rest = if let Some(stripped) = origin.strip_prefix("https://") {
+        stripped
+    } else if let Some(stripped) = origin.strip_prefix("http://") {
+        stripped
+    } else {
+        return false;
+    };
+
+    if rest.is_empty() {
+        return false;
+    }
+
+    if rest.contains('/') || rest.contains('?') || rest.contains('#') || rest.contains('@') {
+        return false;
+    }
+
+    let (host_str, colon_idx) = if rest.starts_with('[') {
+        if let Some(close_bracket) = rest.find(']') {
+            let host = &rest[1..close_bracket];
+            let after_bracket = &rest[close_bracket + 1..];
+            let idx = if after_bracket.is_empty() {
+                None
+            } else if after_bracket.starts_with(':') {
+                Some(close_bracket + 1)
+            } else {
+                return false; // Malformed IPv6 (garbage after closing bracket)
+            };
+            (host, idx)
+        } else {
+            return false;
+        }
+    } else {
+        let idx = rest.rfind(':');
+        let host = if let Some(i) = idx { &rest[..i] } else { rest };
+        (host, idx)
+    };
+
+    if host_str.is_empty() {
+        return false;
+    }
+
+    if let Some(idx) = colon_idx {
+        let port_str = &rest[idx + 1..];
+        if port_str.is_empty() || port_str.parse::<u16>().is_err() {
+            return false;
+        }
+
+        if !rest.starts_with('[') && rest[..idx].contains(':') {
+            return false;
+        }
+    }
+
+    true
+}
+
 #[derive(Clone, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Config {
@@ -714,6 +775,7 @@ pub struct PluginsConfig {
 #[serde(default, deny_unknown_fields)]
 pub struct KoraConfig {
     pub rate_limit: u64,
+    pub cors_allow_origins: Vec<String>,
     pub max_request_body_size: usize,
     pub enabled_methods: EnabledMethods,
     pub auth: AuthConfig,
@@ -740,6 +802,7 @@ impl Default for KoraConfig {
     fn default() -> Self {
         Self {
             rate_limit: 100,
+            cors_allow_origins: vec!["*".to_string()],
             max_request_body_size: DEFAULT_MAX_REQUEST_BODY_SIZE,
             enabled_methods: EnabledMethods::default(),
             auth: AuthConfig::default(),
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -1,8 +1,10 @@
+use http::HeaderValue;
 use serde::{Deserialize, Serialize};
 use solana_sdk::pubkey::Pubkey;
 use spl_token_2022_interface::extension::ExtensionType;
 use std::{fs, path::Path, str::FromStr};
 use toml;
+use url::Url;
 use utoipa::ToSchema;
 
 use crate::{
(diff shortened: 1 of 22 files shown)


crates/lib/src/config.rs
```diff
@@ -26,6 +26,52 @@ use crate::{
...
+        return false;
+    }
+
+    if rest.contains('/') || rest.contains('?') || rest.contains('#') || rest.contains('@') {
+        return false;
+    }
+
+    let colon_idx = if rest.starts_with('[') {
+        if let Some(close_bracket) = rest.find(']') {
+            rest[close_bracket..].find(':').map(|i| close_bracket + i)
```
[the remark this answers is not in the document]
raushan728: hmm i pushed by mistake

crates/lib/src/config.rs
```diff
@@ -26,6 +26,67 @@ use crate::{
...
+        return false;
+    };
+
+    if rest.is_empty() {
+        return false;
+    }
+
+    if rest.contains('/') || rest.contains('?') || rest.contains('#') || rest.contains('@') {
+        return false;
+    }
```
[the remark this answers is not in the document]
raushan728: @dev-jodee Bot keeps finding valid edge cases in the hand-rolled origin validator Its concerns are legit each time, but it feels like we're chasing an unbounded set of edge cases with manual string parsing.

Should I pull in the `url` crate for origin validation, or keep patching the manual parser case-by-case? Wanted to check before adding.

crates/lib/src/validator/config_validator.rs
```diff
@@ -425,6 +426,32 @@ impl ConfigValidator {
...
+                .count();
+
+            if invalid_count == config.kora.cors_allow_origins.len() {
+                warnings.push("cors_allow_origins contains no valid origins - all cross-origin requests will be blocked".to_string());
+            } else if invalid_count > 0 {
+                warnings.push(format!("cors_allow_origins contains {} invalid origin(s) that will be silently filtered out at runtime", invalid_count));
+            }
+        }
+
+        // Validate payment address
```
dev-jodee: lets keep the comments to a minimum, this is redundant

crates/lib/src/validator/config_validator.rs
```diff
@@ -425,6 +426,32 @@ impl ConfigValidator {
             warnings.push("Rate limit is set to 0 - this will block all requests".to_string());
         }
 
+        // Validate CORS origins
+        if config.kora.cors_allow_origins.is_empty() {
+            warnings.push(
+                "cors_allow_origins is empty - all cross-origin requests will be blocked"
+                    .to_string(),
+            );
+        } else if config.kora.cors_allow_origins.iter().any(|o| o == "*") {
```
dev-jodee: can we have constants for *

crates/lib/src/validator/config_validator.rs
```diff
@@ -1400,7 +1428,8 @@ mod tests {
             metrics: MetricsConfig::default(),
         };
 
-        let _ = update_config(config);
+        // Initialize global config
```
dev-jodee: this comment too

crates/lib/src/rpc_server/server.rs
```diff
@@ -109,6 +109,39 @@ fn get_value_by_priority(env_var: &str, config_value: Option<String>) -> Option<
     AuthConfig::resolve_secret(env_var, config_value.as_deref())
 }
 
+fn build_allow_origin(origins: &[String]) -> AllowOrigin {
```
dev-jodee, line 112: this seems like a lot of duplicate from config validator as well? Prob better to centralize those checks

crates/lib/src/config.rs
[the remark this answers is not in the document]
dev-jodee: this make sense ^

dev-jodee (changes requested):

crates/lib/src/validator/config_validator.rs
raushan728: done but what the use case in future?

crates/lib/src/validator/config_validator.rs
raushan728: You are a experienced one, but these kinds of comments help new contributors understand the code quickly

crates/lib/src/rpc_server/server.rs
raushan728: done added a `classify_cors_origins()` helper that both `build_allow_origin()` and the validator now call

crates/lib/src/rpc_server/server.rs
```diff
@@ -110,35 +110,29 @@ fn get_value_by_priority(env_var: &str, config_value: Option<String>) -> Option<
...
+        CorsOriginsClassification::Wildcard { has_redundant } => {
+            if has_redundant {
+                log::warn!("cors_allow_origins contains '*' alongside specific origin(s). The specific origin(s) are redundant and will be silently ignored.");
+            }
+            AllowOrigin::any()
         }
-
-        AllowOrigin::list(parsed_origins)
+        CorsOriginsClassification::AllInvalid { .. } => {
+            log::warn!("None of the provided origin(s) are valid. Must be a valid web origin (e.g., 'https://your-app.com').");
```
dev-jodee: think there's a duplicate log here

crates/lib/src/rpc_server/server.rs
```diff
@@ -566,6 +569,20 @@ mod tests {
         assert!(!debug_str.contains("https://example.com:"));
         assert!(!debug_str.contains("https://[::1]garbage"));
         assert!(!debug_str.contains("https://[::1]garbage:8080"));
+        assert!(!debug_str.contains("https://[not-ipv6]"));
         assert!(debug_str.contains("[]") || debug_str.contains("List([])"));
     }
+
+    #[test]
+    fn test_cors_wildcard_handling() {
+        let wildcard_only = vec!["*".to_string()];
```
dev-jodee: should use the const

crates/lib/src/rpc_server/server.rs
```diff
@@ -566,6 +569,20 @@ mod tests {
...
+        let allow_origin = build_allow_origin(&wildcard_only);
+        let debug_str = format!("{:?}", allow_origin);
+        assert!(debug_str.contains(r#"Const("*")"#));
+
+        let mixed = vec!["*".to_string(), "https://example.com".to_string()];
```
dev-jodee: here as well

crates/lib/src/validator/config_validator.rs
```diff
@@ -427,31 +427,28 @@ impl ConfigValidator {
...
-            } else if invalid_count > 0 {
-                warnings.push(format!("cors_allow_origins contains {} invalid origin(s) that will be silently filtered out at runtime", invalid_count));
+            CorsOriginsClassification::Wildcard { has_redundant } => {
+                if has_redundant {
+                    warnings.push("cors_allow_origins contains '*' alongside specific origin(s). The specific origin(s) are redundant and will be silently ignored.".to_string());
+                }
+            }
+            CorsOriginsClassification::AllInvalid { .. } => {
+                warnings.push("None of the provided origin(s) are valid. Must be a valid web origin (e.g., 'https://your-app.com').".to_string());
+                warnings.push("cors_allow_origins contains no valid origin(s) - all cross-origin requests will be blocked".to_string());
```
dev-jodee: duplicate warning

dev-jodee (changes requested):
couple small things left

crates/lib/src/rpc_server/server.rs
raushan728: It might look a bit repetitive, but both have different purpose - the first one is explain what the operator did with suggestion, and the second one shows the result with reasoning anyway I've combined.

crates/lib/src/config.rs
```diff
@@ -26,6 +28,78 @@ use crate::{
...
+
+    // A valid origin per Fetch spec exactly matches its ascii_serialization.
+    origin == url.origin().ascii_serialization()
+}
+
+pub enum CorsOriginsClassification {
+    Empty,
+    Wildcard { has_redundant: bool },
+    AllInvalid { invalid_origins: Vec<String> },
```
raushan728: nit: `invalid_origins` is unused I originally had it looping to log each origin individually, but switched to a single aggregate warning since here we don't need multiple logs/warnings but forgot to drop the field. Will clean up (make it a unit variant).

raushan728: Done! and sorry for this actually, most of the code was written by AI. It had a loop for the `AllInvalid` case just for log/warning, but I removed it because there is no need for a loop in this case. However, I forgot to remove the data field.

dev-jodee (approved):

PR #663, opened 2026-09-05
feat(rpc): support multiple `API` keys

raushan728:
Migrates `api_key: Option<String>` to `api_keys: Option<Vec<String>>`, with backward-compatible deserialization for existing single-key configs. Compares keys via fixed-length SHA-256 hashes with constant-time comparison across all configured keys, avoiding any length-dependent timing signal.

Note: this is a split-out of #559.

- [x] AI tooling was used
- [ ] No AI tooling was used

Tool and extent: Gemini used for coding diffs based on my specifications; all changes were reviewed and verified by me before committing.

### DIFF
--- a/crates/lib/src/config.rs
+++ b/crates/lib/src/config.rs
@@ -1,8 +1,8 @@
 use http::HeaderValue;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use solana_sdk::pubkey::Pubkey;
 use spl_token_2022_interface::extension::ExtensionType;
-use std::{fs, path::Path, str::FromStr};
+use std::{env, fs, path::Path, str::FromStr};
 use toml;
 use url::Url;
 use utoipa::ToSchema;
@@ -859,10 +859,55 @@ impl Default for LighthouseConfig {
     }
 }
 
+fn deserialize_api_keys<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum ApiKeys {
+        String(String),
+        Vec(Vec<String>),
+    }
+
+    let opt = Option::<ApiKeys>::deserialize(deserializer)?;
+    match opt {
+        Some(ApiKeys::String(s)) => {
+            let trimmed = s.trim();
+            if trimmed.is_empty() {
+                Ok(None)
+            } else {
+                log::warn!("DEPRECATION WARNING: 'api_key' as a single string is deprecated. Please migrate to using 'api_keys' as an array in your configuration.");
+                Ok(Some(vec![trimmed.to_string()]))
+            }
+        }
+        Some(ApiKeys::Vec(v)) => {
+            let filtered: Vec<String> = v
+                .into_iter()
+                .filter_map(|s| {
+                    let trimmed = s.trim();
+                    if trimmed.is_empty() {
+                        None
+                    } else {
+                        Some(trimmed.to_string())
+                    }
+                })
+                .collect();
+            if filtered.is_empty() {
+                Ok(None)
+            } else {
+                Ok(Some(filtered))
+            }
+        }
+        None => Ok(None),
+    }
+}
+
 #[derive(Clone, Serialize, Deserialize, ToSchema)]
 #[serde(default, deny_unknown_fields)]
 pub struct AuthConfig {
-    pub api_key: Option<String>,
+    #[serde(alias = "api_key", deserialize_with = "deserialize_api_keys")]
+    pub api_keys: Option<Vec<String>>,
     pub hmac_secret: Option<String>,
     pub recaptcha_secret: Option<String>,
     pub recaptcha_score_threshold: f64,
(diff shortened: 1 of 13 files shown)


crates/lib/src/rpc_server/auth.rs
```diff
@@ -71,10 +88,29 @@ where
...
+                let mut matched_index = 0;
+
+                for (index, configured_key) in api_keys.iter().enumerate() {
+                    let provided_bytes = provided_key.as_bytes();
+                    let configured_bytes = configured_key.as_bytes();
+                    let matches: bool = if provided_bytes.len() == configured_bytes.len() {
+                        provided_bytes.ct_eq(configured_bytes).into()
+                    } else {
+                        false
+                    };
```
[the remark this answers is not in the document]
raushan728: fixed

crates/lib/src/rpc_server/server.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(test)]
+use crate::tests::usage_limiter_mock::MockUsageTracker as UsageTracker;
```
dev-jodee: why is usage tracker being modified here? Doesn't seem like the mock does anything ?

crates/lib/src/config.rs
```diff
@@ -859,10 +859,55 @@ impl Default for LighthouseConfig {
...
+    }
+
+    let opt = Option::<ApiKeys>::deserialize(deserializer)?;
+    match opt {
+        Some(ApiKeys::String(s)) => {
+            let trimmed = s.trim();
+            if trimmed.is_empty() {
+                Ok(None)
+            } else {
+                log::warn!("DEPRECATION WARNING: 'api_key' as a single string is deprecated. Please migrate to using 'api_keys' as an array in your configuration.");
```
dev-jodee: don't think this will ever print, since loggers are setup after config is loaded (and even config validate doesn't setup the loggers before)

crates/lib/src/rpc_server/auth.rs
```diff
@@ -53,7 +74,7 @@ where
     }
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
-        let api_key = self.api_key.clone();
+        let api_keys = self.api_keys.clone();
```
dev-jodee, line 95: we should hash those once in the new(), instead of hashing everytime

```
pub fn new(api_keys: Vec<String>) -> Self {
        let keys = api_keys.iter().map(|k| {
            let hash = hash_key(k.as_bytes());
            KeyEntry { id: hex::encode(&hash[..4]), hash }
        }).collect();
        Self { keys }
    }
```

crates/lib/src/rpc_server/auth.rs
```diff
@@ -71,10 +92,24 @@ where
...
+            let mut req = Request::from_parts(parts, Body::from(body_bytes));
             if let Some(provided_key) = req.headers().get(X_API_KEY) {
-                // Constant-time comparison prevents timing attacks
-                if provided_key.as_bytes().ct_eq(api_key.as_bytes()).into() {
+                let mut is_valid = false;
+                let mut matched_id = String::new();
+                let provided_hash = hash_key(provided_key.as_bytes());
+
+                for configured_key in api_keys.iter() {
+                    let configured_hash = hash_key(configured_key.as_bytes());
```
dev-jodee: would be cleaner to use iters, something like 

```
let provided_hash = hash_key(provided_key.as_bytes());

if let Some(entry) = keys.iter().find(|e| bool::from(provided_hash.ct_eq(&e.hash))) {
    req.extensions_mut().insert(ClientIdentity(format!("apikey:{}", entry.id)));
    return inner.call(req).await;
}
```

dev-jodee (changes requested):
couple of small changes, looks good !

crates/lib/src/rpc_server/server.rs
raushan728: form https://github.com/solana-foundation/kora/pull/559#discussion_r3546548973

> can't have cfg test flags in the production code, think I mentioned it before, this should be done like we do cache mocking i.e.

meant for the rate-limiting split PR where integration tests actually call `run_rpc_server` with `usage_limit` enabled. It ended up here because most of was carried over from the original branch. Confirmed nothing in this PR's tests calls `run_rpc_server` so removed it - will come back in the PR that actually needs it.

crates/lib/src/config.rs
raushan728: Confirmed - you are right properly surfacing it would mean tracking which deserialization form was used (legacy string vs array) and threading that through to the validator's warnings which needs either a manual deserialize impl with a proxy struct, or a process lifetime static, that felt like more complexity than a startup notice warranted, so I removed the `log::warn` call entirely rather than build that out

crates/lib/src/rpc_server/auth.rs
raushan728: form https://github.com/solana-foundation/kora/pull/559#discussion_r3407948735

> `ct_eq` keeps each individual comparison constant-time, but the early `break` means a key that matches at position 0 exits after one comparison while a key at position 4 runs five. An attacker with many guesses can exploit the response-time difference to determine how many configured keys exist and which ordinal position a valid key occupies, partially undermining the "constant-time comparison prevents timing attacks" guarantee documented on line 97. The fix is to remove the `break` and let all comparisons run, accumulating the result without short-circuiting.

The suggested `.find()` version would reintroduce exactly this `.find()` short-circuits on the first match, so it's the same early-exit problem with a different name. Kept the explicit loop iterating every entry unconditionally, just with hashes now precomputed in new(). Open to a cleaner-looking loop, just can't use `.find()` here.

dev-jodee (approved):

PR #666, opened 2026-09-10
ci: harden disclosure parsing and label cleanup

dev-jodee:
## Summary
- Strip HTML comments across lines with a single `perl -0777` pass. The previous per-line `sed` let a checked disclosure box hidden inside a multiline comment pass the check while GitHub rendered nothing.
- Remove a stale `ai-unreviewed` label when the author turns out to have write access, instead of exiting before cleanup.
- Use an absolute CONTRIBUTING link in the PR template; the relative `../CONTRIBUTING.md` does not resolve once the template is in a PR body.

Ports the Greptile review fixes from solana-foundation/solana-keychain#308 so all Solana Foundation copies of this workflow stay identical. Follow-up to #665.

## Test Plan
- `actionlint` clean, including the SC2001 style finding that the `sed` removal makes go away.
- Verified locally that a `[x]` line inside a multiline `<!-- -->` block is stripped and no longer counted.
- Workflow is now byte-identical (slug-normalized) to the copies in solana-keychain, solana-mcp-official, program-examples, and subscriptions.

## AI disclosure

- [x] AI tooling was used. Tool and extent: Claude Code ported the fixes under direction; the author reviewed every change.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,7 +8,7 @@
 
 ## AI disclosure
 
-Check exactly one. See [CONTRIBUTING.md](../CONTRIBUTING.md#ai-use).
+Check exactly one. See [CONTRIBUTING.md](https://github.com/solana-foundation/kora/blob/main/CONTRIBUTING.md#ai-use).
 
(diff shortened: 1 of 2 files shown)


amilz (approved):
LGTM: hardens disclosure parsing (multiline comment strip) and label cleanup; CI green.

PR #668, opened 2026-09-10
build: prune unused dependencies and drop native-tls

dev-jodee:
## What and why

Follow-up to #664 (cache consolidation). A cold `cargo build --timings` showed Kora's own crates compile in the last ~16 s of a 205 s build, so cold time is dominated by dependencies. This PR is the hygiene half: remove what is not used and stop building two TLS stacks. The aws-lc-sys chain that dominates the critical path comes from solana-keychain and will be handled there.

- Remove the 22 dependencies cargo-machete reports as unused across kora-lib, kora-cli, tests and the devnet-deploy example, plus workspace entries nothing references anymore (env_logger, tracing, hyper, jsonrpsee-core, vaultrs, http-body-util, p256). solana-transaction is a false positive (present to enable the `wincode` feature) and is added to the machete ignore metadata so future runs stay clean.
- Drop `native-tls` from reqwest. rustls with the platform verifier was already in the tree, so this removes openssl, hyper-tls and tokio-native-tls without changing which root store is used.
- Move mockall to dev-dependencies. The production `PriceSource::Mock` oracle becomes a plain `PriceOracle` impl instead of a mockall expectation object, `#[automock]` on `PriceOracle` is gated behind `cfg_attr(test)`, and the unused `TokenInterface` automock is removed.

Cargo.lock goes from 883 to 847 packages. No behavior change for operators. The transfer-hook example program also has an unused `spl-type-length-value` but lives in its own `build-sbf` workspace and is left alone here.

## Testing

- `cargo clippy -- -D warnings`, `cargo fmt --check`
- `cargo test --lib --workspace --exclude tests`: 963 passed
- `cargo check -p tests -p devnet-deploy-paymaster --all-targets`
- `cargo build --release -p kora-cli`
- `cargo-machete --with-metadata .` reports clean

## AI disclosure

Check exactly one. See [CONTRIBUTING.md](../CONTRIBUTING.md#ai-use).

- [x] AI tooling was used. Tool and extent: Claude Code ran the timings and machete analysis and drafted the change; I reviewed the dependency list, the mock oracle rewrite and the TLS change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -344,7 +344,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.119",
- "synstructure 0.13.2",
+ "synstructure",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


amilz (approved):
LGTM: prunes unused deps and drops native-tls; Cargo.lock-only churn, full CI green.

PR #672, opened 2026-09-16
fix: propagate local unit test failures

kriss39:
## What and why

The local Rust and TypeScript unit-test recipes prefix their commands with just's `-` ignore-error marker. A failing test therefore does not propagate a non-zero exit status through `just unit-test`, `just unit-test-ts`, the default recipe, or `just test`, so a red test suite still reports success locally.

Remove the two ignore-error prefixes so test failures fail the invoking recipe and stay visible to contributors and scripts. The test commands themselves are unchanged. CI is not affected: `rust-unit.yml` calls `cargo test --workspace --lib` directly rather than going through `just`.

## Testing

- `git diff --check`
- Verified the diff only removes the two `-` prefixes; the recipe bodies are otherwise identical.
- `just` is not installed on the machine used for this change, so the recipes were not executed locally.

## AI disclosure

Check exactly one. See [CONTRIBUTING.md](https://github.com/solana-foundation/kora/blob/main/CONTRIBUTING.md#ai-use).

- [x] AI tooling was used. Tool and extent: Codex identified the ignore-error prefixes and drafted the two-line justfile change; I reviewed the full diff and wrote the description.


dev-jodee:
@kriss39 signed ocmmits here too pls

dev-jodee (approved):

PR #673, opened 2026-09-16
fix: reject zero balance metrics interval

kriss39:
## What and why

When metrics and fee-payer balance metrics are both enabled, `metrics.fee_payer_balance.expiry_seconds = 0` reaches `tokio::time::interval(Duration::ZERO)` in `BalanceTracker::start_background_tracking`. Tokio panics on a zero period, so the spawned metrics task dies at startup instead of the operator getting an actionable configuration error.

Validate the value in `ConfigValidator` alongside the existing `sign_timeout_seconds` check, and reject zero only while the fee-payer balance collector is active. A zero value in a disabled section still loads as before, so existing inactive configs keep working.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p kora-lib --lib fee_payer_balance_interval` (2 passed)
- `git diff --check`

The two tests cover the rejected case (metrics and collector enabled, interval zero) and the allowed case (collector disabled, interval zero).

## AI disclosure

Check exactly one. See [CONTRIBUTING.md](https://github.com/solana-foundation/kora/blob/main/CONTRIBUTING.md#ai-use).

- [x] AI tooling was used. Tool and extent: Codex located the panic path and drafted the validation check and the two regression tests; I reviewed the full diff and test output and wrote the description.


dev-jodee:
@kriss39 looks good but you need to sign commits

kriss39:
> @kriss39 looks good but you need to sign commits

Done it

dev-jodee (approved):

